Issue #33, opened 2021-03-30
BUG: cgset can't handle `=` in the value


drakenclimber:
See this [comment](https://github.com/libcgroup/libcgroup/issues/12#issuecomment-810431419) in Issue #12.

A user was setting io.max as follows, but the command fails due to the `=` character in the value.
`cgset -r io.max="8:16 wbps=1024" foo`

drakenclimber:
Two additional comments for you @santiagorr

- You're welcome to fix this if you wish.  If not, I should be able to get to it in the next week or two.
- If you want credit for reporting the bug, please populate the following
  `Reported-by: Your Real Name <youremailaddress@arealemail.com>`

drakenclimber:
And, of course, thanks for the help!

santiagorr:
Thanks for confirming it's bug!

> Two additional comments for you @santiagorr
> 
>     * You're welcome to fix this if you wish.  If not, I should be able to get to it in the next week or two.

I think I will be short of cycles to handle this, especially after some recent covid-related government decisions. Please, go ahead if cannot provide anything.

>     * If you want credit for reporting the bug, please populate the following
>       `Reported-by: Your Real Name <youremailaddress@arealemail.com>`
Please use this:
`Reported-by: Santiago Ruano Rincón <santiago.ruano-rincon@imt-atlantique.fr>`

drakenclimber:
Thanks, @santiagorr.  I have issued pull request #35 which should resolve this issue.

PR #35, opened 2021-04-02
Fix issue where cgset can't handle '=; in the value

drakenclimber:
This patchset resolves github issue #33


drakenclimber:
And here's the associated unit test to go with these changes.

PR #45, opened 2021-05-20
Replace __BEGIN_DECLS and __END_DECLS with extern "C".

Hello71:
The macros __BEGIN_DECLS and __END_DECLS are a GNU-ism found in
glibc and uClibc, but not musl.  We replace them by the more general
extern "C" { ... } block exposed only if we have __cplusplus.

Forward ported from a patch by Anthony G. Basile <blueness@gentoo.org>.

### DIFF
--- a/include/libcgroup/config.h
+++ b/include/libcgroup/config.h
@@ -9,7 +9,9 @@
 #include <features.h>
 #endif
 
-__BEGIN_DECLS
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 /**
  * @defgroup group_config 5. Configuration
@@ -128,6 +130,8 @@ int cgroup_config_create_template_group(
  * @}
  * @}
  */
-__END_DECLS
+#ifdef __cplusplus
+} /* extern "C" */
+#endif
 
 #endif /*_LIBCGROUP_CONFIG_H*/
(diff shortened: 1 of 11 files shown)


drakenclimber:
Thanks for the submission, @Hello71.  The changes look good to me.

Could you please add your signed-off-by to the commit?  Now that we're getting more community contributions, I have a review out for contributions guidelines.  The proposed guidelines are available here:
https://github.com/drakenclimber/libcgroup/blob/issues/contributing/CONTRIBUTING.md

Thanks.

Hello71:
done

giani (changes requested):
@Hello71 - If this was a patch written by Anthony - can you please retain is "From" so Anthony gets the appropriate credit.

Hello71:
original patch: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/libcgroup/files/libcgroup-0.41-replace_DECLS.patch

but concept of __BEGIN_DECLS being for internal use only is not new, nor is this fix. ianal but i think sed -i -e 's/__BEGIN_DECLS/#ifdef __cplusplus\nextern "C" {\n#endif/' etc is not copyrightable. i added the comments to fit better with rest of code style, but i don't really care about that, you can remove it if you want, or write your own patch.

drakenclimber:
Thanks @Hello71.  Code has been merged in commit 215974e67a24edf9097507e8707190b6a42c7207

PR #63, opened 2021-09-10
configure.ac: check for the library containing fts_open

### LINKED ISSUE #61: Fails to build with musl
musl doesn't provide `fts.h`[1]  so libcgroup doesn't compile with musl. However, there is a standalone implementation of fts for musl users[2] which can be used.

If configure.ac called `AC_SEARCH_LIBS([fts_open], [fts], [], [AC_MSG_ERROR(...)])` then it would check for `fts_open` in the standard C library, and if it wasn't present check in `libfts`.

[1] https://wiki.musl-libc.org/faq.html
[2] https://github.com/void-linux/musl-fts

rossburton:
The musl C library doesn't provide fts.h[1], so libcgroup doesn't
compile with musl. However, there is a standalone implementation of fts
for musl users[2] which can be used.

Use AC_SEARCH_LIBS to search for fts_open, which will check if it is part
of libc, and if not look in libfts, then set LIBS if needed.

[1] https://wiki.musl-libc.org/faq.html
[2] https://github.com/void-linux/musl-fts

Closes #61.

Signed-off-by: Ross Burton <ross.burton@arm.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -174,6 +174,10 @@ AC_FUNC_REALLOC
 AC_FUNC_STAT
 AC_CHECK_FUNCS([getmntent hasmntopt memset mkdir rmdir strdup])
 
+AC_SEARCH_LIBS([fts_open], [fts], [], [
+  AC_MSG_ERROR([unable to find the fts_open() function])
+])
+
 if test x$with_pam = xtrue; then
 	AC_CHECK_LIB(
 		[pam],


drakenclimber:
Thanks @rossburton.  Merged in commit https://github.com/libcgroup/libcgroup/commit/b28983068a02d94fb616b51b21733b53cc64d210

drakenclimber:
And cherry-picked into the release-2.0 branch in commit c370e44b864171ea7012dc166e14ff2609a1df2b

PR #70, opened 2021-12-02
automake: idiomatically mark test libraries as convenience libraries

rossburton:
Instead of installing then deleting, mark the libraries which are only
used at build time and should not be installed as noinst, so they don't
get installed at all.

Signed-off-by: Ross Burton <ross.burton@arm.com>

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -14,20 +14,17 @@ AM_YFLAGS = -d
 CLEANFILES = lex.c parse.c parse.h
 
 AM_CPPFLAGS = -I$(top_srcdir)/include
-lib_LTLIBRARIES = libcgroup.la libcgroupfortesting.la
+
+lib_LTLIBRARIES = libcgroup.la
 libcgroup_la_SOURCES = parse.h parse.y lex.l api.c config.c libcgroup-internal.h libcgroup.map wrapper.c log.c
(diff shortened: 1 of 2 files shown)


drakenclimber:
> Well that test suite failure is interesting.

Agreed.  I had gone down this path when I first added the funky library deletion stuff.  But I hit the same weird error you are seeing.  Out of expedience, I ended up with what's currently checked in.

I'll try to dig into this some more in the next few days.  I think the `noinst` solution is so much cleaner, and I would much prefer to use it if at all possible.

Thanks again, for the patch @rossburton.

drakenclimber:
What do you think of this small addition to your patch, @rossburton?  I added some `ifndef` checks to cgset.c
https://github.com/drakenclimber/libcgroup/commit/b6831a0fe7340e84a5724f481644d1183198fb8a

All of the unit and functional tests are now passing:
https://github.com/drakenclimber/libcgroup/runs/4409803697

rossburton:
That looks like the right thing to do, yes.

drakenclimber:
Thanks for the patch, @rossburton.  Pushed to main in commit https://github.com/libcgroup/libcgroup/commit/1932bd3c46fc5f3532d6583ccc2ccd4c8d8530a2

drakenclimber:
Pushed to commit c3be09411a40ba09af14fe57df33e1ec3bfb5cd5 in the release-2.0 branch

Issue #64, opened 2021-09-30
BUG: cgroup.threads not supported


bjne:
libcgroup has incorrect support for cgroup2 threads in cgroup_attach_task, where the documentation says
pid = tid, attaching a thread to a cgroup2 group actually adds the thread to cgroup.procs, which causes all
threads of the process to be added

drakenclimber:
Good find, @bjne.  Thanks.  I've added it to my todo list for v2.0.1

PR #72, opened 2021-12-06
api.c: add support for cgroup.threads

kamalesh-babulal:
cgroup_build_tasks_procs_path(), that builds the proc path for moving
the tid, doesn't check for the cgroup type, and appends cgroup.procs for
both cgroup v1/v2.  With cgroup v2, this results in building a wrong
proc path if the cgroup type is threaded. This patch adds support to
check for the cgroup v2 type and build the proc path according to the
below table:
```
cgroup.type     | proc file      | Description                        |
----------------+----------------+------------------------------------+
domain          | cgroup.procs   | regular cgroup, all threads gets   |
                |                | moved to the cgroup.               |
domain threaded | cgroup.procs   | root of threaded subtree, all      |
                |                | threads gets moved to the cgroup.  |
threaded        | cgroup.threads | cgroup under threaded subtree,     |
		|		 | single threads with matching tid,  |
		|		 | gets moved to the cgroup.          |
domain invalid  | N/A            | invalid state of a cgroup under the|
                |                | threaded subtree.                  |
```
Closes: https://github.com/libcgroup/libcgroup/issues/64

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1514,11 +1514,72 @@ char *cg_build_path(const char *name, char *path, const char *type)
 	return path;
 }
 
+static int cgroup_get_cg_type(const char * const path, char * const type,
+			      size_t type_sz)
+{
+	char cg_type_path[FILENAME_MAX];
+	char cg_type[LL_MAX];
+	int len, err = 0;
+	FILE *fp = NULL;
+	struct stat st;
+	int stat_ret;
+
+	snprintf(cg_type_path, FILENAME_MAX, "%scgroup.type", path);
+	/* file cgroup.type, doesn't exist for root cgroup. */
+	stat_ret = stat(cg_type_path, &st);
+	if (stat_ret != 0) {
+		snprintf(type, type_sz, "cgroup.procs");
+		goto out;
+	}
+
+	fp = fopen(cg_type_path, "re");
+	if (!fp) {
+		cgroup_warn("Warning: failed to open file %s: %s\n",
+				cg_type_path, strerror(errno));
+		err = ECGOTHER;
+		goto out;
+	}
+
+	if (fgets(cg_type, LL_MAX, fp) == NULL) {
+		cgroup_warn("Warning: failed to read file %s: %s\n",
+				cg_type_path, strerror(errno));
+		err = ECGOTHER;
+		goto out;
+	}
+
+	len = strlen(cg_type) - 1;
+	/*
+	 * Append cgroup.threads to the path, if the cgroup.type is
+	 * threaded and cgroup.procs for type domain, domain threaded.
+	 * domain type is used for regular cgroup and domain threaded
+	 * for root of threaded cgroup v2 subtree.  Another possible
+	 * type is domain invalid, it's an invalid state, under the
+	 * threaded subtree.
+	 */
+	if (strncmp(cg_type, "domain", len) == 0 ||
+	    strncmp(cg_type, "domain threaded", len) == 0) {
+		snprintf(type, type_sz, "cgroup.procs");
+	} else if (strncmp(cg_type, "threaded", len) == 0) {
+		snprintf(type, type_sz, "cgroup.threads");
+	} else {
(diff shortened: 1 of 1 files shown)


src/api.c
```diff
@@ -1534,8 +1539,55 @@ int cgroup_build_tasks_procs_path(char * const path,
...
+		/*
+		 * Append cgroup.threads to the path, if the cgroup.type is
+		 * threaded and cgroup.procs for type domain, domain threaded.
+		 * domain type is used for regular cgroup and domain threaded
+		 * for root of threaded cgroup v2 subtree.  Another possible
+		 * type is domain invalid, it's an invalid state, under the
+		 * threaded subtree.
+		 */
+		if (strcmp(cg_type, "domain") == 0 ||
+		    strcmp(cg_type, "domain threaded") == 0) {
```
giani, line 1580: So i really do not like using strcmp here - I would rather you use a strncmp -> esp since the string we are compaing against has a fixed size?

kamalesh-babulal: Sure, the reason to use `strcmp()` was to compare the whole string, that is read from `/proc` and there is very little chance of overrun.

Werkov:
Hi. How is the switch of `cgroup.type` to a threaded variant meant to be done? (IIUC, you leave that to an outer entity to switch some subtrees into the threaded mode? (And then only adapt the path creation based on that.))

src/api.c
```diff
@@ -1534,8 +1539,55 @@ int cgroup_build_tasks_procs_path(char * const path,
 		err = 0;
 		break;
 	case CGROUP_V2:
-		strncat(path, "cgroup.procs", path_sz - strlen(path));
 		err = 0;
+		snprintf(cg_type_path, FILENAME_MAX, "%scgroup.type", path);
```
drakenclimber, line 1543: Should we put this reading of `cgroup.type` in its own function?  Seems like this could be helpful for two reasons:
1.  I'm guessing we'll need to read it again elsewhere
2.  It would make it easier to test.  Testing threads in the functional tests might be tedious (or it might not), but it should be trivial to unit test this

drakenclimber:
> Hi. How is the switch of `cgroup.type` to a threaded variant meant to be done? (IIUC, you leave that to an outer entity to switch some subtrees into the threaded mode? (And then only adapt the path creation based on that.))

As of right now you are correct, libcgroup doesn't have a way to read/write `cgroup.type`.  Definitely seems like something we should add.  I added a [github issue](https://github.com/libcgroup/libcgroup/issues/73) to track this.  Thanks, @Werkov

src/api.c
kamalesh-babulal: Agree, will move the reading `cgroup.type` block into a new function and re-spin the patch.  One idea, would be to extend   `process::create_process_in_cgroup()` to create threads using the `threading` module:
```
def create_process_in_cgroup(self, config, controller, cgname,
                                 cgclassify=True, threads=0):
```
if `threads` argument count is more than one, then create `N` threads and write the tid one of them into `cgroup.threads`

kamalesh-babulal:
> Hi. How is the switch of cgroup.type to a threaded variant meant to be done? (IIUC, you leave that to an outer entity to switch some subtrees into the threaded mode? (And then only adapt the path creation based on that.))
@Werkov, thanks for the review. I missed the thought about reading/writing using the libcgroup. @drakenclimber, thanks for adding the patch to read/write.

drakenclimber:
@kamalesh-babulal, do we still need this pull request?  I merged your `threaded` [patchset](https://github.com/libcgroup/libcgroup-tests/commit/abcbe0df3fe2016bf65f37cf4af7e65eff4dc1f2) into the test suite, and it's passing on both v1 and v2 without these changes.

drakenclimber:
Pushed to main in commit https://github.com/libcgroup/libcgroup/commit/4c577f91050b1ad00e5ed2b2abb40fc4978db996.  Pushed to release-2.0 in commit d085455befc3d291064d1438cd4614e1245546a3

Thanks @kamalesh-babulal

Issue #73, opened 2021-12-06
RFE: Add support for reading/writing cgroup.type


drakenclimber:
See pull request #72 and this [comment](https://github.com/libcgroup/libcgroup/pull/72#issuecomment-986799357) in it.

drakenclimber:
Here's a passing test run with a new test that reads/writes `cgroup.type`
https://github.com/drakenclimber/libcgroup/runs/4436976650

And here's the new test:
https://github.com/drakenclimber/libcgroup-tests/commit/9ba430f59d780b5670e2c58bba94fed9266a190e

PR #74, opened 2021-12-06
api.c: Allow reading/writing from cgroup.* files in cgroup v2

drakenclimber:
Allow reading/writing from cgroup.* files (such as cgroup.type
and cgroup.procs) on cgroup v2 filesystems.

Closes: https://github.com/libcgroup/libcgroup/issues/73

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -396,6 +396,17 @@ int cgroup_test_subsys_mounted(const char *name)
 			pthread_rwlock_unlock(&cg_mount_table_lock);
 			return 1;
 		}
+
+		/* The user has likely requested a file like cgroup.type or
+		 * cgroup.procs.  Allow this request as long as there's a
+		 * cgroup v2 controller mounted.
+		 */
+		if (strncmp(name, CGROUP_FILE_PREFIX,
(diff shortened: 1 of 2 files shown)


kamalesh-babulal (approved):

Werkov (commented):
Minor nit: I'd suggest extracting the `"cgroup"` into a constant macro.

Remark: this allows the user to write also into `cgroup.procs`, `cgroup.subtree_control` directly but that should be OK (correct me if wrong) since libcgroup doesn't keep any internal state and it will read the changed values from the kernel when necessary anyway.

drakenclimber:
> Minor nit: I'd suggest extracting the `"cgroup"` into a constant macro.

Agreed.  Thanks.

> Remark: this allows the user to write also into `cgroup.procs`, `cgroup.subtree_control` directly but that should be OK (correct me if wrong) since libcgroup doesn't keep any internal state and it will read the changed values from the kernel when necessary anyway.

Yes, I realized this as I was making the change, and I think it's safe.  But until we have a little better testing in place, I'm going to leave it as an unadvertised feature.

Thanks again for looking over the code.

PR #78, opened 2021-12-14
github: Add Doxyfile and github doxygen action

drakenclimber:
Add a default Doxyfile and a doxygen action.


drakenclimber:
Still some work to do.  Sorry for the noise

PR #79, opened 2021-12-15
github: Add Doxyfile and github doxygen action

drakenclimber:
Add a default Doxyfile and a doxygen action.


drakenclimber:
If we decide we like this, I will add a link to the Doxygen on the libcgroup README.  I don't want to add that now because the branch, `doxygen/main`, won't exist until this is committed.

.github/workflows/continuous-integration.yml
```diff
@@ -34,6 +34,27 @@ jobs:
...
+        submodules: false
+    - uses: mattnotmitt/doxygen-action@v1
+      with:
+        working-directory: "./"
+        doxyfile-path: "./doc/Doxygen"
+    - name: Deploy Doxygen
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_dir: ./html
```
kamalesh-babulal, line 55: That's cool, thanks for working on it. A minor nitpick, should it be deployed under `./doc/html` ?

drakenclimber: Good call.  Thanks!

kamalesh-babulal (approved):

kamalesh-babulal:
Reviewed-by: Kamalesh Babulal <kamalesh.babulal@oracle.com>

PR #80, opened 2021-12-16
Cgroup v1/v2 Abstraction Layer

drakenclimber:
This patchset combines two RFC [1][2] patchsets sent out
earlier.  In addition, a few more v1 <-> v2 mappings have also
been added.

Key changes in this patchset:
* Add cgroup v1 <-> v2 abstraction logic and mappings
* Add two new CLI tools - cgxget and cgxset.  These tools
  can map cgroup v1 requests to a cgroup v2 sysfs and vice
  versa.  (And of course they can handle v1-only and v2-only
  requests)
* Add new C bindings in the tools folder including cgroup_cgxget()
  and cgroup_cgxset()
* Delete old python bindings
* Add new cython python bindings.  This is especially helpful
  for testing the C API changes

The changes to libcgroup are available here:
https://github.com/drakenclimber/libcgroup/tree/issues/cgxget5-pybindings4

The changes to libcgroup-tests are available here:
https://github.com/drakenclimber/libcgroup-tests/tree/issues/cgxget5-pybindings4

Code coverage increased from 51.5% to 53.5%:
https://coveralls.io/builds/45064267

Automated tests are passing and results are here:
https://github.com/drakenclimber/libcgroup/runs/4549430696

[1] https://sourceforge.net/p/libcg/mailman/message/37317688/
[2] https://sourceforge.net/p/libcg/mailman/message/37382047/


drakenclimber:
The test failures are expected because the pull request runs the latest tests on `libcgroup/libcgroup-tests` main branch.  This [pull request](https://github.com/libcgroup/libcgroup-tests/pull/28) to libcgroup-tests fixes the failures as well as adds new tests

src/abstraction-common.c
```diff
@@ -0,0 +1,66 @@
...
+	int ret = 0;
+
+	if (out_value == NULL) {
+		cgroup_err("Error: Invalid parameter to %s\n", __func__);
+		ret = ECGINVAL;
+		goto out;
+	}
+
+	*out_value = strtol(in_str, &endptr, base);
+
```
kamalesh-babulal, line 47: The function looks good, a minor nitpick. Man pages, recommends assigning `errno = 0` before calling `strtol()`

src/abstraction-common.c
```diff
@@ -64,3 +64,47 @@ int cgroup_strtol(const char * const in_str, int base,
...
+	if (strlen(in_value) > 0) {
+		ret = cgroup_strtol(in_value, 10, &out_value);
+		if (ret)
+			goto out;
+
+		/* now scale from the input range to the output range */
+		out_value = out_value * out_dflt_int / in_dflt_int;
+
+		out_value_str = calloc(sizeof(char), OUT_VALUE_STR_LEN);
+		ret = snprintf(out_value_str, OUT_VALUE_STR_LEN, "%ld\n", out_value);
```
kamalesh-babulal, line 94: I guess appending a newline to the control setting is not required while writing the value.

src/tools/cgget.c
```diff
@@ -746,14 +825,30 @@ int main(int argc, char *argv[])
...
 	ret = get_values(cg_list, cg_list_len);
 	if (ret)
 		goto err;
 
+#ifdef CGXGET
+	ret = convert_cgroups(&cg_list, cg_list_len, CGROUP_DISK, version);
+	if (ret)
+		goto err;
+#endif
+
```
kamalesh-babulal, line 870: Should the  `cgxget` and `cgxset` have their own files? though it will share most of the common code.  It makes life easier, though the maintenance overhead will increase. In the commit message, `cgxget -2 cpu.weight MyCgroup` -> `cgxget -2 -r cpu.weight MyCgroup`, it missing `-r` option. Can you also please update other commit messages with `-r` in the examples.

src/abstraction-common.c
```diff
@@ -64,3 +64,47 @@ int cgroup_strtol(const char * const in_str, int base,
...
+
+	if (strlen(in_value) > 0) {
+		ret = cgroup_strtol(in_value, 10, &out_value);
+		if (ret)
+			goto out;
+
+		/* now scale from the input range to the output range */
+		out_value = out_value * out_dflt_int / in_dflt_int;
+
+		out_value_str = calloc(sizeof(char), OUT_VALUE_STR_LEN);
```
kamalesh-babulal, line 93: `calloc` return value needs to be checked for NULL.

src/abstraction-map.c
```diff
@@ -32,13 +32,17 @@
...
 	{cgroup_convert_int, "cpu.shares", (void *)1024, "cpu.weight", (void *)100},
+	{cgroup_convert_passthrough, "cpu.stat", NULL, "cpu.stat", NULL},
 };
 const int cgroup_v1_to_v2_map_sz = sizeof(cgroup_v1_to_v2_map) /
 				   sizeof(cgroup_v1_to_v2_map[0]);
 
 const struct cgroup_abstraction_map cgroup_v2_to_v1_map[] = {
+	/* cpu controller */
 	{cgroup_convert_int, "cpu.weight", (void *)100, "cpu.shares", (void *)1024},
+	{cgroup_convert_passthrough, "cpu.stat", NULL, "cpu.stat", NULL},
```
kamalesh-babulal, line 66: `cpu.stat` is a little special, as its values are in the `multiline`.  This needs to be handled while passing the value of the setting to `cgroup_covnvert_passthrough` hinting it to write into `cgroup_contoller->multiline_value` and not `cgroup_controller->value` or something like that.

drakenclimber: The contents of `cpu.stat` aren't a 1-to-1 mapping from cgroup v1 to v2.  Thus, even passing through the multi-line data still wouldn't be correct.  I'm thinking about marking `cpu.stat` as unmappable for now.

We can then revisit it in a smaller patchset :).  There's also `cgroup_read_stats()` which needs to be made smarter for v1/v2.

Thoughts?

src/tools/cgget.c
drakenclimber: Yeah, I think this is a good plan.  When I first added `cgxget`, there were only a couple small differences between it and `cgget`, so having them in the same file made sense.  But as you noticed, there are several `ifdefs` scattered throughout the code and it's getting ugly.

drakenclimber: And good catch on the erroneous comment :/

src/wrapper.c
```diff
@@ -297,6 +297,33 @@ int cgroup_add_value_bool(struct cgroup_controller *controller,
 	return ret;
 }
 
+int cgroup_remove_value(struct cgroup_controller * const controller,
+			const char * const name)
+{
+	int i;
+
+	for (i = 0; i < controller->index; i++) {
+		if (strcmp(controller->values[i]->name, name) == 0) {
```
kamalesh-babulal, line 306: A minor improvement would be to `free` the `multiline_value`, to avoid a memleak, something like
```       
        for (i = 0; i < controller->index; i++) {
                if (strcmp(controller->values[i]->name, name) == 0) {
+
+                       if (controller->values[i]->multiline_value)
+                               free(controller->values[i]->multiline_value);
+
                        if (i == (controller->index - 1)) {
```

src/tools/cgget.c
```diff
@@ -835,7 +844,12 @@ int main(int argc, char *argv[])
 
 #ifdef CGXGET
 	ret = convert_cgroups(&cg_list, cg_list_len, version, CGROUP_DISK);
-	if (ret)
+	if (ret == ECGNOVERSIONCONVERT && ignore_unmappable)
+		/* The user has specified that we should ignore any errors
+		 * due to being unable to map from v1 to v2 or vice versa
+		 */
+		ret = 0;
```
kamalesh-babulal, line 856: `goto err;` is required here too, otherwise `get_value()` will try to read the non-existing control value and trigger an error ```variable file read failed No such file or directory```

.github/workflows/continuous-integration.yml
```diff
@@ -107,14 +107,12 @@ jobs:
       uses: ./.github/actions/setup-libcgroup
     - name: Run functional tests
       run: |
+        # The cgroup v1 runner hosted by Github Actions doesn't allow
+        # for exclusive cpusets.  Thus, skip the cpuset automated test
         pushd tests/ftests
-        make check
+        ./ftests.py -l 10 -L ftests.log
+        ./ftests.py -l 10 -L ftests-nocontainer.log --skip 37 --no-container
```
kamalesh-babulal, line 115: `make check` would also skip testcase `28`. Should it be skipped here too for no-container test?

configure.ac
```diff
@@ -214,7 +202,6 @@ AC_CONFIG_FILES([Makefile
 	src/daemon/Makefile
 	src/tools/Makefile
 	src/pam/Makefile
-	src/bindings/Makefile
```
kamalesh-babulal, line 217: `src/bindings/` directory also needs to be removed, because the new binding code is placed under `src/python`

src/api.c
```diff
@@ -50,6 +50,12 @@
 #include <linux/un.h>
 #include <grp.h>
 
+const struct cgroup_library_version library_version = {
+	.major = CGROUP_VER_MAJOR,
+	.minor = CGROUP_VER_MINOR,
+	.release = CGROUP_VER_RELEASE,
+};
+
```
kamalesh-babulal, line 58: The `configure.ac`, has the following values for the library versions:
```# set library version, soname is libcgroup.so.MAJOR
AC_SUBST(LIBRARY_VERSION_MAJOR, 1)
AC_SUBST(LIBRARY_VERSION_MINOR, 0)
AC_SUBST(LIBRARY_VERSION_RELEASE, 42)
```
 Do those need to be updated to `2.0.0`?

src/abstraction-map.c
kamalesh-babulal: Agree, for now, it can be marked as un-mappable.

src/wrapper.c
drakenclimber: Excellent find!  You are spot on.

drakenclimber: In fact, the problem is even more insidious.  The `memmove()` just below this line is wrong.  It's supposed to move the array of pointers forward one slot in the array.  But it's actually moving the contents of one pointer (and potentially garbage after it) forward into the contents of the pointer that's been removed.  So it's a memory leak, potentially a security hole (by copying wrong data), and potentially a memory corruption by overwriting the memory after the value that's been removed.  Ugh!!

src/api.c
drakenclimber: Yes, but I think we should mimic what libseccomp has done here.  In the main branch's [configure.ac](https://github.com/seccomp/libseccomp/blob/main/configure.ac#L22), the version is set to `0.0.0`.  This way anything built with the main branch will have a strange version number that should signify to a user that it's not an official release.  In the release branch, then we can set the version to the appropriate number.  As you pointed out, `2.0.0` or whatever.

configure.ac
drakenclimber: I'm not seeing `src/bindings/` in anything committed into github (related to this patchset).  Is it just an artifact laying around in your sandbox from an old build?

src/tools/cgget.c
drakenclimber: I don't think so.  The way I did it, `ignore_mappable` throws an error and doesn't populate that setting.  But there could be other valid settings that do need to be converted.

Unless I'm missing something entirely... :)

src/wrapper.c
drakenclimber: But I have a fix for it... Coming soon :)

.github/workflows/continuous-integration.yml
drakenclimber: We should probably update `make check`.  Test 028 no longer exists.

drakenclimber:
I updated the [tests](https://github.com/libcgroup/libcgroup-tests/pull/28) and this branch.  I believe all of your review comments have been addressed, @kamalesh-babulal.   Thanks so much!

Here's a [branch](https://github.com/drakenclimber/libcgroup/tree/wip/cgxget3) that shows all of the tests are passing.

.github/workflows/continuous-integration.yml
kamalesh-babulal: Agreed, I created a[ PR ](https://github.com/libcgroup/libcgroup-tests/pull/33) to remove the `skip` argument.

configure.ac
kamalesh-babulal: Ah !! sorry for the noise, it was from the old build.

src/api.c
kamalesh-babulal: Interesting idea ! we should do the same.

src/tools/cgget.c
drakenclimber: Sorry for being slow on the uptake.  You are right - this is a problem because cgxget will display every setting if the `cgroup_controller struct` doesn't have any settings within it.

I've pushed a proposed fix.  Here are the [two](https://github.com/drakenclimber/libcgroup/commit/11f6606827872bcebfcd7a6e3a5391eff7ae2e33#diff-5e93f3166e4154f7d03bcbb4228ecbf5f10fee8e82d724d830c8f1a224061504R235) [key](https://github.com/drakenclimber/libcgroup/commit/11f6606827872bcebfcd7a6e3a5391eff7ae2e33#diff-5e93f3166e4154f7d03bcbb4228ecbf5f10fee8e82d724d830c8f1a224061504R288) parts of the proposed fix.  The tests all pass, but I'm not explicitly testing this case.  We should add this.

drakenclimber: After thinking more about it, this code won't work.  `cgc` will be a stale pointer to the freed memory.  Bummer
```
			if (!cgc)
				/*
				 * The converted controller had no settings
				 * and was removed.  It's up to us to manage
				 * the controller count
				 */
				out_cgroup->index--;
```

drakenclimber: Fixed now.  I passed `struct cgroup_controller ** const out_cgc` to `convert_controller()` and NULL out the the pointer if the controller is empty.

drakenclimber:
Yet another update.  Passing tests are available [here](https://github.com/drakenclimber/libcgroup/tree/wip/cgxget3).

src/tools/cgget.c
drakenclimber: I added a test for this explicit case.  Here's the code coverage showing it got hit on both a v1 and v2 system:
https://coveralls.io/builds/46171147/source?filename=src%2Fabstraction-common.c#L220

kamalesh-babulal (approved):

kamalesh-babulal:
> I updated the https://github.com/libcgroup/libcgroup-tests/pull/28 and this branch. I believe all of your review comments have been addressed, @kamalesh-babulal. Thanks so much!

@drakenclimber Yes, the latest patch series takes care of all the comments.  Thank you so much of working on it :)

drakenclimber:
Merged into main.  Thanks all!

PR #85, opened 2022-01-12
ENH: Add Security Features

drakenclimber:
* Add a security document that outlines the process for handling security-related vulnerabilities
* Add a continuous integration step that utilizes the CodeQL scanning tool.  It [currently reports](https://github.com/drakenclimber/libcgroup/security/code-scanning?query=is%3Aopen+branch%3Aissues%2Fcodeql)
  four issues in Googlemock (which we'll obviously ignore) and one issue in libcgroup itself.  Not
  sure yet if this is an issue we should address or not.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -34,6 +34,22 @@ jobs:
           echo "Cleaning up previous run"
           rm -rf "${{ github.workspace }}"
 
+  codeql:
+    name: CodeQL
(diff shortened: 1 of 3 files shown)


drakenclimber:
Security status is prominently featured in github via the "Security" tab at the top of the libcgroup home page:
https://github.com/libcgroup/libcgroup/security

kamalesh-babulal (approved):

Issue #84, opened 2021-12-26
BUG: pam_cgroup simply doesn't work with cgroup2


0n-s:
pam_cgroup, at least under a cgroup2-only environment (maybe cgroup1 as well, I haven't tested), simply doesn't seem to work at all. It still acts as if it succeeded.

Tracing a random sshd instance for every file it opens, there are only 3:
```
25666 sshd                3   0 /proc/cgroups
25666 sshd                5   0 /proc/mounts
25666 sshd                7   0 /sys/fs/cgroup/cgroup.controllers
```
(nothing else cgroup or libcgroup-related)
For whatever reason cgrules.conf is never loaded. Despite this, it claims it changed the cgroup successfully (with `debug` flag):
```
Dec 26 13:49:12 h pam_cgroup(sshd:session): Changed cgroup for process 25666  with username [redacted].
```
This is not true, as we can see from the just logged in session:
```
[redacted]@h $ cat /proc/self/cgroup
0::/
```
This failure also doesn't result in the login failing, even though in my PAM configuration pam_cgroup is not optional but required.

(tested with libcgroup v2.0)

drakenclimber:
Thanks for the bug report, @0n-s.

The pam module hasn't been a priority for the current users of libcgroup, so we have not emphasized development or testing in this area.  I'm uncertain how much the libcgroup/pam module is being used, as this is the first issue reported against it even though v2.0 has been out 7+ month.  Hmmmm....

drakenclimber:
I'm going to assign this to v2.0.1.  I'm not certain if it will be fixed in that release, but I would like us to make a decision on how to proceed with pam by that release.

Thanks again for the report!

lockywolf:
The pam module is a great thing to have. Cgrulesengd is a hack anyway.

I think the Slackware community has seen similar problems with cgroupv1 (and had to rely on cgrulesengd), but sometimes people are just too lazy or shy to report bugs :(.

drakenclimber:
> The pam module is a great thing to have. Cgrulesengd is a hack anyway.

Won't argue with the cgrules comment :)

> I think the Slackware community has seen similar problems with cgroupv1 (and had to rely on cgrulesengd), but sometimes people are just too lazy or shy to report bugs :(.

That's good to know.  Thank you.

drakenclimber:
Thanks for the comments, @0n-s and @lockywolf.  I have merged a fix into `main` and `release-2.0`.  Let me know if you have any issues.

[...]

PR #87, opened 2022-01-14
api.c: cache cgrules with CGFLAG_USECACHE flag

kamalesh-babulal:
Services using pam_cgroup.so plugin, asks to match a rule, in the
cached list of rules already constructed from /etc/cgrules.conf.
This works well with active cgrulesengd but in the instances where
its disabled, the rules are not read and cached by default. Fix,
this is by reloading and caching the rules, when called with
CGFLAG_USECACHE flag.

Fixes: https://github.com/libcgroup/libcgroup/issues/84

Reported-by: Kent Overstreet <kent.overstreet@gmail.com>

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -3877,6 +3877,20 @@ int cgroup_change_cgroup_flags(uid_t uid, gid_t gid,
 	}
 
 	/*
+	 * User had asked to find the matching rule (if one exist) in the
+	 * cached rules  but the list might be empty due to the inactive
+	 * cgrulesengd. Lets emulate its behaviour of caching the rules
+	 * by reloading the rules from the configuration file.
+	 */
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):
Thanks, @kamalesh-babulal.  I did some basic tests on these changes on cgroup v1 on Ubuntu20 and cgroup v2 on Ubuntu20.  Everything looked good.

kamalesh-babulal:
@drakenclimber Thanks for the review and testing.

drakenclimber:
Thanks, @kamalesh-babulal.

Fixed in main in commit 80a0bd4b8873df38060309e775d609dcc219531f.  Fixed in release-2.0 in commit b7275a0fcb85be77c4dfdc953f16adfcaa558723

PR #91, opened 2022-02-02
configure: set version numbers for releases only

kamalesh-babulal:
Let's set the version for releases only and set the development/main
branch version to 0.0.0, that differentiate between development and
releases.

Suggested-by: Tom Hromatka <tom.hromatka@oracle.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,7 +15,7 @@
 AC_PREREQ([2.69])
 
 # In following section update all occurences of version, including soname
-AC_INIT([libcgroup],[2.0])
+AC_INIT([libcgroup],[0.0.0])
 
 AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_MACRO_DIRS([m4])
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Looks good to me.  I really hate how code coverage numbers in coveralls is all over the board.  The overall coverage remained the same (which is obviously correct), but in one of the intermediate steps coverage went down 2.1%.  Argh.

drakenclimber:
I re-ran the tests, and coverage is now what I expect.  But the interesting part is the the "summary" coverage job is listed near the end.  During the failing run, it was in the middle.  I wonder if parallelization isn't setup quite right.

kamalesh-babulal:
> I re-ran the tests, and coverage is now what I expect. But the interesting part is the the "summary" coverage job is listed near the end. During the failing run, it was in the middle. I wonder if parallelization isn't setup quite right.

Sorry, I did not wait until the code coverage results were published. I never thought it would make a difference to the coverage and  2.1% decrement is huge. I am guessing the #92 did the magic?

drakenclimber:
> > I re-ran the tests, and coverage is now what I expect. But the interesting part is the the "summary" coverage job is listed near the end. During the failing run, it was in the middle. I wonder if parallelization isn't setup quite right.
> 
> Sorry, I did not wait until the code coverage results were published. I never thought it would make a difference to the coverage and 2.1% decrement is huge. I am guessing the #92 did the magic?

No worries.  It's obviously not your fault and something is wrong with the ordering of the jobs.  My _hope_ is that pull #92 fixed it.  We'll see :)

PR #94, opened 2022-02-07
Add support for creating an empty cgroup v2 cgroup

drakenclimber:
The goal of this patchset is to allow a user to create and delete an empty cgroup v2 cgroup:
cgcreate like so:
```
cgcreate -g :foo
cgdelete -g :foo
```

Automated tests are passing:
https://github.com/drakenclimber/libcgroup/runs/5101568130

Code coverage is at 52.75%.  All new lines (except error handling) were covered:
https://coveralls.io/builds/46312484

New tests are available here:
https://github.com/libcgroup/libcgroup-tests/pull/34


drakenclimber:
(Force push to remove the hacks to use the temporary test branch.)

src/api.c
```diff
@@ -2455,88 +2436,119 @@ int cgroup_create_cgroup(struct cgroup *cgroup, int ignore_ownership)
...
-					cgroup->tasks_uid, cgroup->tasks_gid,
-					cgroup->task_fperm);
-			if (error)
-				goto err;
-		}
-		free(base);
-		base = NULL;
 	}
+	free(base);
+	base = NULL;
```
kamalesh-babulal, line 2518: @drakenclimber The patchset looks great, I gave it a try and it works as expected.  A few minor nitpicks, the above `free()` and `base = NULL` can be removed, it's free below.

src/api.c
```diff
@@ -2436,29 +2440,37 @@ static int _cgroup_create_cgroup(const struct cgroup * const cgroup,
 	fts_path[1] = NULL;
 	path = fts_path[0];
 
-	if (!cg_build_path(cgroup->name, path, controller->name))
-		goto err;
+	if (controller) {
+		if (!cg_build_path(cgroup->name, path, controller->name))
+			return 0;
```
kamalesh-babulal, line 2445: return here will cause a memory leak of `fts_path`, it should be replaced with `goto err;` and return in the else block below to needs to be `goto err`.

drakenclimber: D'oh.  Good catch.

kamalesh-babulal (approved):

PR #104, opened 2022-02-28
Fixes for GCC warnings

kamalesh-babulal:
While compiling on Fedora-34, which ships `GCC-11` as the
default compiler triggers a couple of `strncpy()` related build
warnings.  This patch series attempts to fix those warnings
and keep the GCC happy.


drakenclimber (approved):

drakenclimber:
Thanks, @kamalesh-babulal.  Pushed to main in commits b275d89b489234f3dc2eef9784873a7903508a5c and 23068e8f107ed3de4bde36c15f5ff32c5ffcc835

PR #106, opened 2022-03-09
[RFC] src/* - introduce coding style

kamalesh-babulal:
This is the patch series is the first attempt to bring up the `src/*.[ch]` files,
close to the coding style recommended by Linux Kernel's `/scripts/checkpatch.pl`.
It also introduces reverse Xmas declaration for local variables and re-orders the
headers files in the following order:
1. Header files that declare the functions, are defined in this file.
2. Other header files from the libcgroup.
3. Standard header file.

not all recommendations from `checkpatch.pl` are applicable to us, so this series
adds `.checkpatch.conf` at the top-level of the source directory, where the following
recommendations are ignored:
1. `--no-tree` - This hints that we are running it over the non-kernel tree.
2. `--ignore SPDX_LICENSE_TAG` - Linux Kernel excepts, the first line of the source
    to have SPDX License tag and it does not apply to us.
3. `--ignore USE_NEGATIVE_ERRNO` - The error code are positive in our source files,
    so ignore this error.
4. `--max-line-length=80` - Stick to 80 column limit for now.
5. `--ignore SSCANF_TO_KSTRTO` - ignore the recommendation of `kstr*` functions,
    those are implemented in the Kernel.

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -1,0 +1,15 @@
+# There is no Kernel tree available, run without it.
+--no-tree
+
+# Ignore the SPDX License tag, that's supposed to be the first line in
+# Linux Kernel sources.
+--ignore SPDX_LICENSE_TAG
+
+# We do not follow the negative error numbering, so ignore it.
+--ignore USE_NEGATIVE_ERRNO
+
+# Follow the 80-character limit.
+--max-line-length=80
+
+# Kernel specific kstr* functions recommendation doesn't apply to us.
+--ignore SSCANF_TO_KSTRTO
--- a/src/abstraction-common.c
+++ b/src/abstraction-common.c
@@ -19,20 +19,21 @@
  * along with this library; if not, see <http://www.gnu.org/licenses>.
  */
 
-#include <libcgroup.h>
-#include <libcgroup-internal.h>
-
-#include <errno.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-#include <unistd.h>
-
 #include "abstraction-common.h"
 #include "abstraction-map.h"
 
+#include <libcgroup.h>
+#include <libcgroup-internal.h>
+
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+#include <stdio.h>
+
+
 int cgroup_strtol(const char * const in_str, int base,
-		  long int * const out_value)
+		  long * const out_value)
 {
 	char *endptr = NULL;
 	int ret = 0;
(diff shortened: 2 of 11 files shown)


drakenclimber (approved):

drakenclimber:
Changes look good to me.  Thanks, @kamalesh-babulal.

Should we invoke `checkpatch.pl` in our GitHub Actions?  We could curl or wget the latest from the Linux kernel, then run it as part of the continuous integration.  Thoughts?

drakenclimber:
Changes look good to me.  Thanks, @kamalesh-babulal.

Should we invoke `checkpatch.pl` in our GitHub Actions?  We could curl or wget the latest from the Linux kernel, then run it as part of the continuous integration.  Thoughts?

kamalesh-babulal:
> Should we invoke checkpatch.pl in our GitHub Actions? We could curl or wget the latest from the Linux kernel, then run it as part of the continuous integration. Thoughts?

Thanks for the review @drakenclimber.  I was thinking of fixing the recommendations by `checkpatch.pl` on our sources files before integrating the check into `github-actions` so that the pull requests do not fail because checkpatch.pl is unhappy.  There is one `github-actions` available https://github.com/webispy/checkpatch-action, that we would use.

giani:
mind also putting in the SPDX tags in the files? it is all LGPL 2.1

kamalesh-babulal:
> mind also putting in the SPDX tags in the files? it is all LGPL 2.1

Thanks for the review @giani, in fact, I have a tree that converts the license boilerplate into `SPDX License tag` https://github.com/kamalesh-babulal/libcgroup/tree/spdx-license.  While we are at this, one question I had was regarding the https://github.com/libcgroup/libcgroup/blob/main/src/pam/pam_cgroup.c, is it safe to remove the LGPL-2.1 boilerplate and add `SPDX-License-Identifier: LGPL-2.1-only` to it and keep the original copyright notice intact?

kamalesh-babulal:
Another question regarding the licensing is, how do we go about on the source files without any license by the original author?

drakenclimber:
> I was thinking of fixing the recommendations by `checkpatch.pl` on our sources files before integrating the check into `github-actions` so that the pull requests do not fail because checkpatch.pl is unhappy.

Are all of the checkpatch warnings/errors fixed once these patches are applied?  If not, what's left?

kamalesh-babulal:
>    I was thinking of fixing the recommendations by checkpatch.pl on our sources files before integrating the check into github-actions so that the pull requests do not fail because checkpatch.pl is unhappy.

> Are all of the checkpatch warnings/errors fixed once these patches are applied? If not, what's left?

Ah, my bad. Sorry, I should have explicitly mentioned in the description, that this PR only cleans up the source files under `./src/*`.  The remaining ones are from `src/tools/*`, `src/pam/*`, `src/daemon/*`, `include/*`, `include/libcgroup/*` and `samples/c/*`, there are another 36+ patches those fixes all of the remaining source files, I was planning to do different PR for the individual directory for ease of reviewing or we could just push it here. I do not have strong opinions on it.

PR #107, opened 2022-03-10
log: add log level string prefix to logging functions

kamalesh-babulal:
Usage of cgroup_{err, warn, info} functions, expects the log level to be
prefixed to the message like: cgroup_warn("Warning: ...").  This patch
removes this need to prefix and automatically adds the desired prefix to
the logger function, so the usage becomes cgroup_warn("...").  This
helps in readability and avoiding typos in setting the right log level
string to the message.

### DIFF
--- a/src/libcgroup-internal.h
+++ b/src/libcgroup-internal.h
@@ -76,9 +76,9 @@ extern "C" {
 
 #define CGROUP_FILE_PREFIX	"cgroup"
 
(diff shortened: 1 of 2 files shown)


kamalesh-babulal:
The idea is to remove `Error, Info, Warning` keywords from the existing `cgroup_log*` statements if this looks good.

drakenclimber:
Merged into main in commit 97a0e195ddd780795674e0c1672cdfb5c3e86a0e.  Thanks, @kamalesh-babulal

PR #108, opened 2022-03-11
[RFC] Adopt SPDX License tag

kamalesh-babulal:
Adopt SPDX license tag for all the source files, those already have
LGPL 2.1 bolierplate in them.  Adopting SPDX license helps the
compliance tools to determine the license and also helps in reducing the
repetitive license boilerplate across source files.

### DIFF
--- a/src/tools/cgconfig.c
+++ b/src/tools/cgconfig.c
@@ -1,17 +1,9 @@
-
-/*
+// SPDX-License-Identifier: LGPL-2.1-only
+/**
  * Copyright IBM Corporation. 2007
  *
  * Authors:	Dhaval Giani <dhaval@linux.vnet.ibm.com>
  * 		Balbir Singh <balbir@linux.vnet.ibm.com>
- *
- * This program is free software; you can redistribute it and/or modify it
- * under the terms of version 2.1 of the GNU Lesser General Public License
- * as published by the Free Software Foundation.
- *
- * This program is distributed in the hope that it would be useful, but
- * WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  *
  * Code initiated and designed by Dhaval Giani. All faults are most likely
  * his mistake.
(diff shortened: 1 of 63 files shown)


kamalesh-babulal:
@giani asked for picking up of SPDX License tag for the source files, in #106.  This PR is not complete, we need to decide on, how to go about the files that do not have licenses already.

src/tools/cgconfig.c
```diff
@@ -1,18 +1,10 @@
-
-/*
+// SPDX-License-Identifier: LGPL-2.1-only
+/**
```
drakenclimber, line 2: (This is a note to myself as much as anything.)

Having the copyright header be a doxygen comment added a bunch of noise to libseccomp's doxygen output.
https://github.com/seccomp/libseccomp/pull/366/commits/f2548f2fb21d8e2274c372eb48deb6a6f9f134d8#r785130185

Leave it as you have it, but we may want to look more closely at our doxygen sometime.

drakenclimber (approved):

giani:
Thanks @kamalesh-babulal . This looks good to me.

For files that do not have an explicit license, they are implied to be LGPL-2.1 only (because that's what the project is). I would explicitly license them as LGPL-2.1

giani (approved):

kamalesh-babulal:
> For files that do not have an explicit license, they are implied to be LGPL-2.1 only (because that's what the project is). I would explicitly license them as LGPL-2.1

Thanks, @giani for the review/suggestion.

src/tools/cgconfig.c
kamalesh-babulal: @drakenclimber Thanks for the review.  Interestingly, Doxygen doesn't recognize Copyright. Agree, we might need to audit it.

kamalesh-babulal:
**V2 Changes**
- Resolve the conflicts against the current main, to help in a clean merge.
- Additional commit that adds `SPDX-License` identifier to files missing license information.

drakenclimber:
Pushed to main in commits 7e30070e882f and 2bfb5500af22.  Thanks @kamalesh-babulal

PR #109, opened 2022-03-11
samples/c: drop the test suites

kamalesh-babulal:
samples/c contains test cases, those do serve as good examples on how to
use some APIs, but there are also very elaborate test cases, which are
test suites in themselves.  Let's drop these examples and allow users to
look at gunit/ftests for how to contribute to test cases.  This patch
also modifies the Makefile to help users to build the examples, in case
they want to give it a quick try.


drakenclimber (approved):

drakenclimber:
Pushed to main in commit f1f4f64cc68356f41af95f72c1205724d8830e2c.  Thanks, @kamalesh-babulal

PR #111, opened 2022-03-15
api.c: API to list mount points of a cgroup type

kamalesh-babulal:
Currently, there is no easy way for a user to list cgroup mount points,
one way to acquire required information is by manually reading the
`/proc/mounts` and parsing the information they are interested in.

Add a new API:
`cgroup_list_mount_points(cg_version_t version, char ***mnts)`

where the first argument is either CGROUP_V1 or CGROUP_V2, which
specifies the mount point types and the second argument is a char**,
that will hold the mount point paths of the cgroup version specified in
the first argument. Note that the mnts pointers are supposed to free'd.
mnts pointers.

```
$ cat get_mount.c

int main(void)
{
        enum cg_version_t t = CGROUP_V2;
        char **mount_paths;
        int i = 0;
        int ret;

        cgroup_init();

         ret = cgroup_list_mount_points(t, &mount_paths);
         if (ret != 0) {
                 fprintf(stderr, "Failed to get mount points\n");
                 return ret;
        }

        while (mount_paths[i]) {
                fprintf(stderr, "%s\n", mount_paths[i]);
                free(mount_paths[i]);
                i++;
        }

        return 0;
}

$ gcc -o get_mount get_mount.c -lcgroup
```

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -93,6 +93,9 @@ __thread char *cg_namespace_table[CG_CONTROLLER_MAX];
 pthread_rwlock_t cg_mount_table_lock = PTHREAD_RWLOCK_INITIALIZER;
 struct cg_mount_table_s cg_mount_table[CG_CONTROLLER_MAX];
 
+/* Cgroup v2 mount paths, with empty controllers */
+struct cg_mount_point *cg_cgroup_v2_empty_mount_paths;
+
 const char * const cgroup_strerror_codes[] = {
 	"Cgroup is not compiled in",
 	"Cgroup is not mounted",
@@ -1221,7 +1224,32 @@ STATIC int cgroup_process_v2_mnt(struct mntent *ent, int *mnt_tbl_idx)
 
 	ret_c = fgets(line, LL_MAX, fp);
 	if (ret_c == NULL) {
+		struct cg_mount_point *tmp, *t;
+
 		ret = ECGEOF;
+
+		tmp = malloc(sizeof(struct cg_mount_point));
+		if (tmp == NULL) {
+			last_errno = errno;
+			ret = ECGOTHER;
+			goto out;
+		}
+
+		strncpy(tmp->path, cg_cgroup_v2_mount_path,
+			sizeof(tmp->path) - 1);
+		tmp->path[sizeof(tmp->path)-1] = '\0';
+		tmp->next = NULL;
+
+		t = cg_cgroup_v2_empty_mount_paths;
+		if (t == NULL) {
+			cg_cgroup_v2_empty_mount_paths = tmp;
+			goto out;
+		}
+
+		while(t->next != NULL)
+			t = t->next;
+		t->next = tmp;
+
 		goto out;
 	}
 
@@ -1292,6 +1320,8 @@ static void cgroup_free_cg_mount_table(void)
 
 	memset(&cg_mount_table, 0, sizeof(cg_mount_table));
 	memset(&cg_cgroup_v2_mount_path, 0, sizeof(cg_cgroup_v2_mount_path));
+	memset(&cg_cgroup_v2_empty_mount_paths, 0,
+	       sizeof(cg_cgroup_v2_empty_mount_paths));
 }
 
 /*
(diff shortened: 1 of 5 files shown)


kamalesh-babulal:
**V2 Changes**
- Handle the empty Cgroup V2 mount paths.

src/api.c
```diff
@@ -1237,7 +1240,30 @@ STATIC int cgroup_process_v2_mnt(struct mntent *ent, int *mnt_tbl_idx)
 
 	ret_c = fgets(line, LL_MAX, fp);
 	if (ret_c == NULL) {
+		struct cg_mount_point *tmp, *t;
+
 		ret = ECGEOF;
+
+		tmp = malloc(sizeof(struct cg_mount_point));
```
drakenclimber, line 1247: Where is `tmp` freed?

src/api.c
```diff
@@ -5907,6 +5933,91 @@ int cgroup_get_controller_version(const char * const controller,
...
+
+	/*
+	 * Cgroup v2 can be mounted without any controller and these mount
+	 * paths are not part of the cg_mount_table.  Check and append them
+	 * to mnt_paths.
+	 */
+	if (cg_cgroup_v2_empty_mount_paths) {
+		mount_point = cg_cgroup_v2_empty_mount_paths;
+		while (mount_point) {
+			mnt_paths[idx] = strdup(mount_point->path);
```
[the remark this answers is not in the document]
kamalesh-babulal: Thanks for the review @drakenclimber, yes it's possible. We can either increase the array size or increase it the day someone ever complains about it?

kamalesh-babulal:
**V3 Changes**
- Increased the mount points size to 10% of what `sysctl fs.mount-max`
- Included overflow checks, to no exceed allocated mount point size, as suggested by @drakenclimber

drakenclimber:
What if we did something like this?
https://github.com/drakenclimber/libcgroup/commit/7e6f04f2766273012945b65f526c3eced1549bb6#diff-180c1c48614a941a92369affdf11a798b80ca8db3ad6d2d7c79338fd3d7a64d4R5933

Then we avoid having to pre-allocate an array of 10000 which won't be used 99.9% of the time.  But we're still protected if a system is setup that way.

kamalesh-babulal:
> What if we did something like this?
https://github.com/drakenclimber/libcgroup/commit/7e6f04f2766273012945b65f526c3eced1549bb6#diff-180c1c48614a941a92369affdf11a798b80ca8db3ad6d2d7c79338fd3d7a64d4R5933
Then we avoid having to pre-allocate an array of 10000 which won't be used 99.9% of the time. But we're still protected if a system is setup that way.

Thanks, @drakenclimber, it looks neat to abstract the bound check and allocation.  Inspired by it, I have re-spined a patch with no static allocation.  Please have a look and let me know your thoughts.

kamalesh-babulal:
**V4 Changes**
- Move from static to dynamic allocation for mount_paths array, it removes the extra checks for overflow and makes it future-proof.

kamalesh-babulal:
**V5 Changes**
- Added python binding for the `cgroup_list_mount_points`.
- Fixed a compilation warning of `mnt_paths` usage.

src/python/libcgroup.pyx
```diff
@@ -277,6 +277,31 @@ cdef class Cgroup:
...
+
+        Arguments:
+        version - It specifies the cgroup version
+
+        Description:
+        Parse the /proc/mounts and list the cgroup mount points matching the
+        version
+
+        Note:
+        List cgroup mount points the specified version"""
```
drakenclimber, line 291: This note seems redundant with the main comment.  Would it make sense to remove it?

src/python/libcgroup.pyx
```diff
@@ -277,6 +277,31 @@ cdef class Cgroup:
...
+        cdef char **a
+
+        mount_points = []
+        ret = cgroup.cgroup_list_mount_points(version, &a)
+        if ret is not 0:
+            raise RuntimeError("cgroup_list_mount_points failed: {}".format(ret));
+
+        i = 0
+        while a[i]:
+            mount_points.append(<bytes>(a[i]).decode("utf-8"))
```
drakenclimber, line 301: Do we want a `bytes` object or do we want a `str` object?

src/python/libcgroup.pyx
kamalesh-babulal: @drakenclimber Thanks for the review. Agree, will remove the redundant comment.

src/python/libcgroup.pyx
kamalesh-babulal: The latter `str`, will change it.

kamalesh-babulal:
**V6 Changes**
- Addressed the python bindings review comments from @drakenclimber 
- Rebased it to the current main branch

drakenclimber (approved):

drakenclimber:
The change looks good to me, @kamalesh-babulal.  Let's get some tests around this in a subsequent commit, so that we can get the coverage numbers back up.

drakenclimber:
Pushed to main in commits 90f586ef5d3f and 66485b72950d

PR #113, opened 2022-03-19
Makefile: fix samples/c build dependency

kamalesh-babulal:
If the user chooses to build the samples/c programs, it will fail to
build, with the error:

config.status: creating samples/c/Makefile
  CC       setuid.o
make: *** No rule to make target '../../src/.libs/libcgroup.la', needed by 'setuid'.  Stop.

it is due to the dependency on libcgroup.la, which is currently built
after
samples. Fix it by moving the samples directory to be built after src.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,6 @@
 AUTOMAKE_OPTIONS = foreign
 ACLOCAL_AMFLAGS= -I m4
-SUBDIRS = dist doc include samples scripts src tests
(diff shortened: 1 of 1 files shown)


bsingharora (approved):

drakenclimber (approved):

drakenclimber:
Thanks, @kamalesh-babulal.  Pushed to main in commit d0ac7d75a31057896a2af45a6ac76f308eab3867

PR #117, opened 2022-03-21
Support simpler cleanup in functions

bsingharora:
Add support for a defer() style function, which cleans up
the memory associated with an auto/local variable when it
goes out of scope. The assumption is that this code is
compiled on a compile with support for the cleanup attribute.

Changes in the patch:
1. Add util.h with support for defer
2. Add common_charptr_cleanup() function to be called for
   freeing char pointers (largely strings in our case containing
   paths)
3. Modify src/api.c to use the new framework

TODO:
1. Support other types of defer functions
2. Support fd closure in cleanup
3. Implement these changes for other files

Testing:
Very light, just compiled and ran simple commands.

Signed-off-by: Balbir singh <bsingharora@gmail.com>


drakenclimber:
Thanks for the submission, @bsingharora.  I admit that I want to think about this some more.

As another reference point, here's where systemd added similar logic (10 years ago!). They did it slightly differently.
https://github.com/systemd/systemd/commit/2fbe635a83a79f8889afec421ae3990ea106fb91

drakenclimber:
@bsingharora Does this work with non-gcc toolchains?

bsingharora:
> @bsingharora Does this work with non-gcc toolchains?

No.. it does not, clang supports it. It is sort of in the commit log "The assumption is that this code is
compiled on a compile with support for the cleanup attribute"

bsingharora:
> Thanks for the submission, @bsingharora. I admit that I want to think about this some more.
> 
> As another reference point, here's where systemd added similar logic (10 years ago!). They did it slightly differently. [systemd/systemd@2fbe635](https://github.com/systemd/systemd/commit/2fbe635a83a79f8889afec421ae3990ea106fb91)

Yep, I like their implementation, I find it interesting that they deref and cast to a void **. I need to implement similar things on fd close (like I mentioned in the log).

PR #121, opened 2022-03-25
make bash scripts shellcheck complaint

kamalesh-babulal:
This patch series fixes the bash scripts shell check complaints.  It allows
exceptions at a couple of sites for rules `SC1090`  and  `SC2034`.


drakenclimber (approved):

drakenclimber:

PR #124, opened 2022-03-29
[RFC] Reduce the size of CG_CONTROLLER_MAX

kamalesh-babulal:
CG_CONTROLLER_MAX defines the maximum possible controllers, those might
be available on the cgroup version, and it's defined as 100.  This size
is way too large in comparison to the number of controllers currently
supported in the Linux Kernel (v5.17):
```
$ grep ^SUBSYS include/linux/cgroup_subsys.h |wc -l
15
```
this translates to over-allocation of memory, which is never used.
So let's reduce it to a count of the current list of the available
controllers with additional room for 10 more controllers, in case
they get added in the future, reducing it to 25 from the current
definition of 100.

**With CG_CONTROLLER_MAX 100**
```
$ size libcgroup.so.0.0.0
   text    data     bss     dec     hex filename
 150123    3608 2474136 2627867  28191b libcgroup.so.0.0.0
```
**With CG_CONTROLLER_MAX 25**
```
$ size libcgroup.so.0.0.0
   text    data     bss     dec     hex filename
 150027    3608  625552  779187   be3b3 libcgroup.so.0.0.0
```
the size of bss is reduced by ~75% and an overall decrease in size
by ~70%.


Werkov:
Beware that there can be an arbitrary number of named v1 hieararchies.
(Actually, by quick look at `cgroup_process_v1_mnt()`, I'm not sure it can handle gracefully overflowing `CG_CONTROLLER_MAX`. But that's a problem with 100 or lower anyway. I'm not against the reduction.

BTW your commit message reports on value of 25 but the patch modifies it to 30.

kamalesh-babulal:
> Beware that there can be an arbitrary number of named v1 hieararchies.
(Actually, by quick look at cgroup_process_v1_mnt(), I'm not sure it can handle gracefully overflowing CG_CONTROLLER_MAX. But that's a problem with 100 or lower anyway. I'm not against the reduction.)

@Werkov Thanks for the review.  Thats true,  we can have more than 100 named hierarchies and in those cases, where  hierarchies are > 100,  we are segfaulting:
```
$ sudo ./src/tools/cgcreate -g cpu:foo;for i in $(seq 1 100); do sudo mkdir /mount$i; sudo mount -t cgroup -o none,name=name$i none /mount$i; done;sudo ./src/tools/cgget -g cpu:foo
Segmentation fault
```

Let me try to work on solutions on how to gracefully handle such cases.

> BTW your commit message reports on value of 25 but the patch modifies it to 30.

My bad, I did not update the right value.

drakenclimber:
> Let me try to work on solutions on how to gracefully handle such cases.

Based on this comment, I'm going to mark this as DRAFT until we get a few things sorted out.  @kamalesh-babulal, feel free to remove the DRAFT label once you feel this is ready

kamalesh-babulal:
> Based on this comment, I'm going to mark this as DRAFT until we get a few things sorted out. @kamalesh-babulal, feel free to remove the DRAFT label once you feel this is ready

Thanks, @drakenclimber. Sure, will do that

kamalesh-babulal:
This patchset will require changes to gunit test cases available at https://github.com/kamalesh-babulal/libcgroup-tests/tree/gunit-dynamic-cg_mount_table for CI to pass.  The passing result tests are available at https://github.com/kamalesh-babulal/libcgroup/actions/runs/2226130436.

Issue #125, opened 2022-03-30
Deletion of v2 cgroup with no controllers fails


kamalesh-babulal:
A cgroup v2 created with no controllers, fails to delete, following is a reproducer
(extension of `samples/c/empty_cgroup_v2.c`):
```
#include <libcgroup.h>

#include <stdlib.h>
#include <stdio.h>

#define CGRP_NAME "empty_cgrp"

int main(int argc, char **argv)
{
        struct cgroup *cgroup = NULL;
        int ret = 0;

        ret = cgroup_init();
        if (ret) {
                fprintf(stderr, "cgroup_init failed\n");
                exit(1);
        }

        cgroup = cgroup_new_cgroup(CGRP_NAME);
        if (!cgroup) {
                fprintf(stderr, "Failed to allocate cgroup %s\n", CGRP_NAME);
                exit(1);
        }

        ret = cgroup_create_cgroup(cgroup, 0);
        if (ret)
                fprintf(stderr, "Failed to create cgroup %s\n", CGRP_NAME);

        ret = cgroup_delete_cgroup_ext(cgroup, CGFLAG_DELETE_RECURSIVE);
        if (ret)
                fprintf(stderr, "Failed to delete cgroup %s\n", CGRP_NAME);

        cgroup_free(&cgroup);

        return ret;
}
```

PR #126, opened 2022-03-31
[RFC] api.c: add support for empty cgroup v2 deletion

kamalesh-babulal:
Add support to recongize empty cgroup v2 with no controllers attached to
it in cgroup_delete_cgroup_ext(), that gets called to remove a cgroup.

In the deletion path, cgroup_build_tasks_procs_path() builds the path
of the cgroup for re-assignment/movement of tasks from the cgroup that's
getting deleted, teach it to recongize the empty controller as cgroup
V2.

Fixes: https://github.com/libcgroup/libcgroup/issues/125

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -3053,9 +3053,11 @@ int cgroup_delete_cgroup_ext(struct cgroup *cgroup, int flags)
 {
 	int first_error = 0, first_errno = 0;
 	char parent_path[FILENAME_MAX];
+	char *controller_name = NULL;
 	FILE *parent_tasks = NULL;
 	char *parent_name = NULL;
 	int delete_group = 1;
+	int empty_cgroup = 0;
 	int i, ret;
 
 	if (!cgroup_initialized)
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
**V2 Changes**
- Added check before `controller_name` assignment.

src/api.c
```diff
@@ -3068,19 +3070,31 @@ int cgroup_delete_cgroup_ext(struct cgroup *cgroup, int flags)
...
 	}
 
-	/* Remove the group from all controllers. */
-	for (i = 0; i < cgroup->index; i++) {
+	/*
+	 * Remove the group from all controllers and in the case of cgroup
+	 * with no controllers, perform all actions of a single controller.
+	 */
+	for (i = 0;
+	     empty_cgroup > 0 || i < cgroup->index; i++, empty_cgroup--) {
```
drakenclimber, line 3087: nifty :)

drakenclimber (approved):

drakenclimber:
Pushed to main in commit ca55c929e8074e630f3fd2eded96542b88ef1ad5.  Thanks, @kamalesh-babulal

src/api.c
kamalesh-babulal: Thanks for the review @drakenclimber

Issue #127, opened 2022-03-31
cgroup v1 - cgroup deletion fails with mount shared by controllers 


kamalesh-babulal:
While debugging #125, there was another issue discovered while deleting a
cgroup created on mount-point shared between two controllers. For example:
```
$ mount|grep -i cpuacct
cgroup on /sys/fs/cgroup/cpu,cpuacct type cgroup (rw,nosuid,nodev,noexec,relatime,cpu,cpuacct)

$ ll /sys/fs/cgroup/|grep -i cpuacct
lrwxrwxrwx  1 root root  11 Mar 31 08:24 cpu -> cpu,cpuacct/
lrwxrwxrwx  1 root root  11 Mar 31 08:24 cpuacct -> cpu,cpuacct/
dr-xr-xr-x  5 root root   0 Mar 31 08:24 cpu,cpuacct/
```
both `cpu` and `cpuacct` are mounted on `/sys/fs/cgroup/cpu,cpuacct` and individually they are
symlinks to the same mountpoint.  There is no issue when both the controllers are added to the
cgroup during the creation but when trying to recursively delete the created cgroup the deletion
fails due to missing cgroup.  The reason being the cgroup on the shared mount point is deleted
when accessed by the first of the shared controller and goes missing for the second controller,
while it tries to delete it.

```
$ cat test.c

#include <stdio.h>
#include <libcgroup.h>

int main(int argc, char **argv)
{
        struct cgroup *cgroup_name = NULL;
        int ret;

        ret = cgroup_init();
        if (ret) {
                fprintf(stderr, "failed cgroup_init\n");
                return (1);
        }

        cgroup_name = cgroup_new_cgroup("mygroup");
        if (!cgroup_name) {
                fprintf(stderr, "failed new_Cgroup\n");
                return (1);
        }

        if (!cgroup_add_controller(cgroup_name, "cpuset")) {
                fprintf(stderr, "failed add_controller cpuset\n");
                return (1);
        }

        if (!cgroup_add_controller(cgroup_name, "cpu")) {
                fprintf(stderr, "failed add_controller cpu\n");
                return (1);
        }

        if (!cgroup_add_controller(cgroup_name, "cpuacct")) {
                fprintf(stderr, "failed add_controller cpuacct\n");
                return (1);
        }
        if (!cgroup_add_controller(cgroup_name, "memory")) {
                fprintf(stderr, "failed add_controller memory\n");
                return (1);
        }

[...]

PR #128, opened 2022-04-01
[RFC] api.c: fix deleting cgroup created on shared mnt point

kamalesh-babulal:
cgroup v1 allows mounting of two or more controllers on a single
mount point.  This fails currently when trying to delete the cgroup
recursively for all the mounted controllers because the cgroup is
already deleted by the first controller of the share mount point and
the controllers then on complaint about missing cgroup path.

Fix this issue by checking if the cgroup controllers share mount point
and ignore the error if the file doesn't exist, similar to ignoring if
the file exists during creation.

Fixes: https://github.com/libcgroup/libcgroup/issues/127

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1073,11 +1073,34 @@ int cg_add_duplicate_mount(struct cg_mount_table_s *item, const char *path)
 	return 0;
 }
 
+/*
+ * Tries to find if any controller in cg_mount_table have already mounted on
+ * the mount_path and if mounted sets the matching controller idx share_mnt
+ * flag and Return 1 or 0 otherwise.
+ */
+static int cgroup_set_cg_mnt_tbl_shared_mnt(char *mount_path, int *mnt_tbl_idx)
+{
+	int i, shared_mnt = 0;
+
+	/* Check if controllers share mount points */
+	for  (i = 0; i < *mnt_tbl_idx; i++) {
+		if (strncmp(mount_path, cg_mount_table[i].mount.path,
+			    FILENAME_MAX) == 0) {
+			cg_mount_table[i].shared_mnt = 1;
+			shared_mnt = 1;
+			break;
+		}
+	}
+
+	return shared_mnt;
+}
+
 static void cgroup_cg_mount_table_append(const char *name,
 					 const char *mount_path,
 					 enum cg_version_t version,
 					 int *mnt_tbl_idx,
-					 const char *mnt_opts)
+					 const char *mnt_opts,
+					 int shared_mnt)
 {
 	int i = *mnt_tbl_idx;
 
@@ -1087,8 +1110,10 @@ static void cgroup_cg_mount_table_append(const char *name,
 	strncpy(cg_mount_table[i].mount.path, mount_path, FILENAME_MAX);
 	cg_mount_table[i].mount.path[FILENAME_MAX-1] = '\0';
 
+	cg_mount_table[i].shared_mnt = shared_mnt;
 	cg_mount_table[i].version = version;
 	cg_mount_table[i].mount.next = NULL;
+
 	cgroup_dbg("Found cgroup option %s, count %d\n", mnt_opts, i);
 
 	(*mnt_tbl_idx)++;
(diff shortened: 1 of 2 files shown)


src/api.c
```diff
@@ -2714,6 +2714,46 @@ static int cgroup_get_parent_name(struct cgroup *cgroup, char **parent)
...
+	int i, j;
+
+	pthread_rwlock_rdlock(&cg_mount_table_lock);
+
+	for (i = 0; cg_mount_table[i].name[0] != '\0'; i++) {
+		if (strncmp(cg_mount_table[i].name, controller,
+			    sizeof(cg_mount_table[i].name)) == 0) {
+
+			controller_mnt_point =
+				strdup(cg_mount_table[i].mount.path);
```
drakenclimber, line 2734: Do you need to `strdup` here?  What if you just grab the address of the mount path?
```
controller_mnt_point = &cg_mount_table[i].mount.path;
```

src/api.c
```diff
@@ -2714,6 +2714,46 @@ static int cgroup_get_parent_name(struct cgroup *cgroup, char **parent)
...
+			if (controller_mnt_point == NULL)
+				goto error;
+
+			break;
+		}
+	}
+
+	for (j = 0; controller_mnt_point && j < i; j++) {
```
drakenclimber, line 2742: Thinking out loud here.... what if the shared mount point is later in `cg_mount_table` than `i`?  Will this code still find it?

src/api.c
```diff
@@ -2714,6 +2714,46 @@ static int cgroup_get_parent_name(struct cgroup *cgroup, char **parent)
...
+			    sizeof(cg_mount_table[i].name)) == 0) {
+			ret = 1;
+			break;
+		}
+	}
+
+	free(controller_mnt_point);
+
+error:
+	pthread_rwlock_rdlock(&cg_mount_table_lock);
```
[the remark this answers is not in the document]
kamalesh-babulal: Thanks for the review @drakenclimber :( that's right it should have been unlock.

src/api.c
kamalesh-babulal: Agree, that should be enough.

src/api.c
```diff
@@ -2755,6 +2795,10 @@ static int cgroup_find_parent(struct cgroup *cgroup, char *controller,
 	cgroup_dbg("parent's name is %s\n", parent_path);
 
 	if (stat(child_path, &stat_child) < 0) {
+		if (is_cgrp_ctrl_mnt_shared(cgroup, controller)) {
```
drakenclimber, line 2798: More thinking out loud here... What if we kept track of all of the deleted paths?  Then we could compare the `child_path` with previously deleted paths.  If it's already been deleted, we can probably just `continue` and move on.

Thoughts?

src/api.c
```diff
@@ -2714,6 +2714,46 @@ static int cgroup_get_parent_name(struct cgroup *cgroup, char **parent)
 	return ret;
 }
 
+/*
+ * Checks if the cgroup's controller shares the mount point with any other
+ * controller of the cgroup. Returns 1 if shared or 0.
+ */
+static int is_cgrp_ctrl_mnt_shared(struct cgroup *cgroup, char *controller)
```
drakenclimber, line 2721: I think if we go with this function, we really need to unit test it.  Should be fairly easy to do, and it should allow us to catch corner cases.

src/api.c
kamalesh-babulal: This function gets called when the `cgroup` is missing.  If the shared mount point is the after `i` and this is the first of the controllers sharing the mount point, we need to worry because something went wrong.

src/api.c
kamalesh-babulal: I did think about it and was biased towards the current implementation, due to the frequency of such use cases.  The check gets performed on error,  to confirm if the directory was deleted by the previous controller or if it's a real bug.  Another approach I was thinking of is adding `cg_mount_point->shared` flag, which gets set during `cgroup_init` and we would quickly check if it's a shared mount point, again in this approach we are assuming that a first of the controller sharing the mount point deleted it.

drakenclimber: The `shared` flag is an interesting idea

kamalesh-babulal: I have implemented the `shared_mnt` flag idea at https://github.com/kamalesh-babulal/libcgroup/commit/6e156aa1c1e195a60e811e1f4a541620356d76a1 please let me know if you like it better than the current approach.

kamalesh-babulal:
**V2 Changes**
-  Implemented `shared_mnt` flag approach.

drakenclimber (approved):

drakenclimber:
Pushed to main in commit 7f9134019910a6357ce0ab778caa50d957ef9ce3.  Thanks, @kamalesh-babulal

PR #130, opened 2022-04-05
[RFC] add support to attach tid to an empty cgroup v2

kamalesh-babulal:
Currently, `cgroup_attach_task_pid()` checks for enabled controllers
before attaching the given tid to it, but in the case of empty cgroup,
it simply ignores the attaching due to no controllers available, a.k.a.
`cgroup->index` is 0.  Add support to recognize the empty controller
cgroups.   Also, always return true in `cgroupv2_controller_enabled()`
for cgroups with no controllers attached.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1858,6 +1858,8 @@ err:
 int cgroup_attach_task_pid(struct cgroup *cgroup, pid_t tid)
 {
 	char path[FILENAME_MAX] = {0};
+	char *controller_name;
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commits fb8a49dc5551165716ff368df593a91b4f422e0d and c8e694c6f0a1330cdcd11b3601a2a5c58fc00df9

PR #136, opened 2022-05-01
api.c: fix segfault in cgroup_populate_mount_points()

kamalesh-babulal:
In cgroup_populate_mount_points(),  cgroup v1/v2 mount points get read
from /proc/mounts and populated into cg_mount_table[].  The size of
cg_mount_table[] is set to CG_CONTROLLER_MAX, that's defined as 100 and
if the system has more than CG_CONTROLLER_MAX, unique mount points, it
will segfault while processing them.  Fix this by checking, the mount
point count after processing every mount entry and bailout in case it
reaches CG_CONTROLLER_MAX.

The issue can be reproduced using, the following simple bash commands on
cgroup v1:
```
1. sudo for i in $(seq 0 100); do sudo mkdir /name$i; done
2. sudo for i in $(seq 0 86); do sudo mount -t cgroup -o none,name=named$i
none /name$i; done
3. sudo cgget -g <controller>:<existing cgroup name>
```

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1466,6 +1466,9 @@ static int cgroup_populate_mount_points(char *controllers[CG_CONTROLLER_MAX])
 			if (ret)
 				goto err;
 
+			if (found_mnt >= CG_CONTROLLER_MAX)
+				break;
+
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
pushed to main in commit 9ce90c7edd281b359f67ee15b67fd28083ed4167.  Thanks, @kamalesh-babulal

PR #137, opened 2022-05-04
api.c: fix segfault while processing cgroup mount points

kamalesh-babulal:
The cgroup v1 and cgroup v2 mount points get read from /proc/mounts in
cgroup_process_v1_mnt()/cgroup_process_v2_mnt() respectively and gets
populated into cg_mount_table[].  The size of cg_mount_table[] is set
to CG_CONTROLLER_MAX, that's defined as 100 and if the system has more
than CG_CONTROLLER_MAX, unique mount points, it will segfault while
processing them.  Fix this by checking, the mount point count after
processing every mount entry and bailout in case it reaches
CG_CONTROLLER_MAX.

The issue can be reproduced using, following simple bash commands on
cgroup v1:
1. sudo for i in $(seq 0 100); do sudo mkdir /name$i; done
2. sudo for i in $(seq 0 100); do sudo mount -t cgroup -o
   none,name=named$i none /name$i; done
3. sudo cgget -g <controller>:<existing cgroup name>

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1116,6 +1116,9 @@ STATIC int cgroup_process_v1_mnt(char *controllers[], struct mntent *ent,
 		cgroup_dbg("Found cgroup option %s, count %d\n",
 			ent->mnt_opts, *mnt_tbl_idx);
 		(*mnt_tbl_idx)++;
+
+		if (*mnt_tbl_idx >= CG_CONTROLLER_MAX)
+			goto out;
 	}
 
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Thanks, @kamalesh-babulal.  Pushed to `release-2.0` in commit daada06e1c606364d5dd20a717b74f91686ec98d

PR #138, opened 2022-05-06
[RFC] api.c: resize controller name to 32 from 4096 bytes

kamalesh-babulal:
Linux kernel warns on the cgroup controller name length greater than 32
bytes during the cgroup initialization in cgroup_init_early().  Adopt
the same size of 32 bytes (including the null byte) for controller name
in the struct cg_controller and struct cg_mount_table_ and struct
cg_mount_table_s.  This resize from FILENAME_MAX (4096 bytes) to a mere
32 bytes, reduces the libcgroup.so library:
```
Without the patch:
------------------
$ size -d src/.libs/libcgroup.so
   text    data     bss     dec     hex filename
 123841    3600 2476532 2603973  27bbc5 src/.libs/libcgroup.so

With the patch:
---------------
$ size -d src/.libs/libcgroup.so
   text    data     bss     dec     hex filename
 123569    3600 1257332 1384501  152035 src/.libs/libcgroup.so
```
also replace the index FILENAME_MAX with CONTROL_NAMELEN_MAX at the
struct cg_controller::name and struct cg_mount_table_s::name references.


kamalesh-babulal:
gunit test cases are also need updating, the passing action results https://github.com/kamalesh-babulal/libcgroup/actions/runs/2282673089

PR #140, opened 2022-05-13
lssubsys, lscgroup - support status update in help/man page

kamalesh-babulal:
lssubsys and lscgroup tools need major rework to support the
cgroups v2.  For now add the support status as notes to both
the help option and the man page of the tools, stating it's supported\
on cgroups v1 only.

### DIFF
--- a/doc/man/lscgroup.1
+++ b/doc/man/lscgroup.1
@@ -40,6 +40,9 @@ in subgroup student
 .B lscgroup
 list all cgroups which in all hierarchies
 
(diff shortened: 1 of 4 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to `main` and the `release-2.0` branch.

I added a `v2.0.3` milestone.  Not sure yet if it will come to pass, but it seemed best to be prepared.

PR #149, opened 2022-05-20
Fix uninitialized variable warnings reported by Coverity

kamalesh-babulal:
This patch series fixes the uninitialized variable warnings reported
by Coverity  across: `tools/cgset, tools/lssubsys, tools/cgsnapshot`
and `tools/cgxset`.


drakenclimber:
Thanks, @kamalesh-babulal.  All commits pushed to `main`.

PR #151, opened 2022-05-23
configure: introduce --enable-tests/--disable-tests

kamalesh-babulal:
There are scenarios where running tests as part of the make check is
not desirable, like rpmbuild.  The build systems might not have the
environment to run the tests and might fail to build the packages.

This patch introduces, --enable-tests as a configure option, which
is set by default and can be disabled using the --disable-tests flag.
When disabled, the configure will skip creating the Makefiles for the
tests/* directories and in the top-level Makefile the SUBDIRS omits
tests. For example, consider the rpmbuild case to build rpm with
--disable-tests passed:
1. ./bootstrap.sh
2. ./configure ./configure --enable-opaque-hierarchy="name=systemd"
   --disable-tests
3. make check
4. make dist
5. cp libcgroup-3.0.0.tar.gz ~/rpmbuild/SOURCES
6. append --disable-tests to configure in libcgroup.spec
7. rpmbuild -ba libcgroup.spec

this will skip running the test/*, during the build, avoiding unexpected
failures

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,9 @@
 AUTOMAKE_OPTIONS = foreign
 ACLOCAL_AMFLAGS= -I m4
-SUBDIRS = dist doc include scripts src tests samples
+SUBDIRS = dist doc include samples scripts src
+if WITH_TESTS
+SUBDIRS += tests
+endif
 
 EXTRA_DIST = README_daemon libcgroup.doxyfile README_systemd CONTRIBUTING.md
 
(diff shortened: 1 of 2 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 17b3a2d612cc8747dcf916b0cf29775cc88f4a23.  Thanks, @kamalesh-babulal

PR #154, opened 2022-05-25
tools/cgget: fix coverity warning about double free

kamalesh-babulal:
Fix double free warning, reported by Coverity tool:

CID 258297 (#1 of 1): Double free (USE_AFTER_FREE). double_free:
Calling closedir frees pointer dir which has already been freed.

As per the man page, the closedir(), closes the directory stream
associated with the dirp, but is ambiguous about if dirp is set to NULL
or not. Coverity answers that with the above report, that dirp is not
NULL and hence the double free.

Fixes: fea1ab8b45d7 ("tools/cgget: fix coverity warning about resource leak")

### DIFF
--- a/src/tools/cgget.c
+++ b/src/tools/cgget.c
@@ -625,10 +625,8 @@ static int fill_empty_controller(struct cgroup * const cg,
 			continue;
 
 		ret = cgroup_fill_cgc(ctrl_dir, cg, cgc, i);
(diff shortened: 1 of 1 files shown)


drakenclimber:
Pushed to `main` in commit 703bbb8554ce882eb461e16c389d20a356f095de.  Thanks, @kamalesh-babulal

PR #159, opened 2022-05-27
api: fix coverity warning about uninitialized variable

kamalesh-babulal:
CID 258269 (#1 of 1): Uninitialized scalar variable (UNINIT).
uninit_use: Using uninitialized value version.

In _cgroup_create_cgroup(), the (cg_version_t)version is uninitialized
and might be read the version in false path. It worked until now
because the version is assigned in the true path and given that its enum
the checks are not bounded by range, but rather specific.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -2430,7 +2430,7 @@ err:
  */
 int cgroup_create_cgroup(struct cgroup *cgroup, int ignore_ownership)
 {
-	enum cg_version_t version;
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to `release-2.0` in commit e61a3ca6561068ccd31bb816477a8936a4b021e0.  Thanks, @kamalesh-babulal

PR #162, opened 2022-05-31
config: add missing ret value check from getgrnam_r()

kamalesh-babulal:
Fix Unchecked return values from library, reported by Coverity tool:

CID 258287 (#1 of 1): Unchecked return value from library
(CHECKED_RETURN).
check_return: Calling getgrnam_r(value, group, buffer, 20480UL,
&group_buffer) without checking return value. This library function may

CID 258303 (#1 of 1): Unchecked return value from library
(CHECKED_RETURN).
check_return: Calling getgrnam_r(value, group, buffer, 20480UL,
&group_buffer) without checking return value. This library function may

Coverity expects us to check for return value from getgrnam_r(), instead
of the current for group_buffer != NULL.  Which is right, let's make
Coverity happy by moving the check to return value.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -365,10 +365,8 @@ int config_group_task_perm(char *perm_type, char *value, int flag)
 			if (!group)
 				goto group_task_error;
 
-			getgrnam_r(value, group, buffer,
-				   CGROUP_BUFFER_LEN, &group_buffer);
-
-			if (group_buffer == NULL) {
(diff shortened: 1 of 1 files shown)


drakenclimber:
Pushed to `main` in commit 18cc60ddf3cba088fce3062099d8ee5fca5341a0.  Thanks, @kamalesh-babulal

PR #163, opened 2022-05-31
config: add missing ret value check from getgrnam_r() [v2.0]

kamalesh-babulal:
Fix Unchecked return values from library, reported by Coverity tool:

CID 258287 (#1 of 1): Unchecked return value from library
(CHECKED_RETURN).
check_return: Calling getgrnam_r(value, group, buffer, 20480UL,
&group_buffer) without checking return value. This library function may

CID 258303 (#1 of 1): Unchecked return value from library
(CHECKED_RETURN).
check_return: Calling getgrnam_r(value, group, buffer, 20480UL,
&group_buffer) without checking return value. This library function may

Coverity expects us to check for return value from getgrnam_r(), instead
of the current for group_buffer != NULL.  Which is right, let's make
Coverity happy by moving the check to return value.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -370,10 +370,8 @@ int config_group_task_perm(char *perm_type, char *value, int flag)
 			if (!group)
 				goto group_task_error;
 
-			getgrnam_r(value, group, buffer,
-					CGROUP_BUFFER_LEN, &group_buffer);
-
-			if (group_buffer == NULL) {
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to `release-2.0` in commit 5e06dc20ab01317f20e61746d9b6baea4e9f8c9c.  Thanks, @kamalesh-babulal

PR #166, opened 2022-06-02
api.c: search_and_append_mnt_path() use strncpy()

kamalesh-babulal:
Fix copy into fixed size buffer warning, reported by Coverity tool:

CID 258279 (#1 of 1): Copy into fixed size buffer (STRING_OVERFLOW).
fixed_size_dest: You might overrun the 4096-character fixed-size string
mnt_point->path by copying path without checking the length.
10. parameter_as_source: Note: This defect has an elevated risk
because the source argument is a parameter of the current function.

fix this issue by using strnpy(), in place of strcpy().

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -6085,7 +6085,9 @@ static int search_and_append_mnt_path(struct cg_mount_point **mount_point,
 		return ECGOTHER;
 	}
 
-	strcpy(mnt_point->path, path);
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to `main` in commit aa6d6301581f86759291b65704e91a31cb100cb4.  Thanks, @kamalesh-babulal

Issue #168, opened 2022-06-09
BUG: ambiguous controller name matches by using "hasmntopt"


Hackerl:
https://github.com/libcgroup/libcgroup/blob/v2.0.2/src/api.c#L1078

calling "hasmntopt" to determine if the controller name exists in "mntopt", may cause errors because of  "hasmntopt" only match substring.

```
Found cgroup option rw,nosuid,nodev,noexec,relatime,xattr,release_agent=/lib/systemd/systemd-cgroups-agent,name=systemd, count 0
found cpuset in rw,nosuid,nodev,noexec,relatime,cpuset
Found cgroup option rw,nosuid,nodev,noexec,relatime,cpuset, count 1
found cpu in rw,nosuid,nodev,noexec,relatime,cpuset
Found cgroup option rw,nosuid,nodev,noexec,relatime,cpuset, count 2
found devices in rw,nosuid,nodev,noexec,relatime,devices
Found cgroup option rw,nosuid,nodev,noexec,relatime,devices, count 3
found perf_event in rw,nosuid,nodev,noexec,relatime,perf_event
Found cgroup option rw,nosuid,nodev,noexec,relatime,perf_event, count 4
found net_cls in rw,nosuid,nodev,noexec,relatime,net_cls,net_prio
Found cgroup option rw,nosuid,nodev,noexec,relatime,net_cls,net_prio, count 5
found net_prio in rw,nosuid,nodev,noexec,relatime,net_cls,net_prio
Found cgroup option rw,nosuid,nodev,noexec,relatime,net_cls,net_prio, count 6
found pids in rw,nosuid,nodev,noexec,relatime,pids
Found cgroup option rw,nosuid,nodev,noexec,relatime,pids, count 7
found cpu in rw,nosuid,nodev,noexec,relatime,cpu,cpuacct
controller cpu,cpuacct is already mounted on /sys/fs/cgroup/cpuset
found cpuacct in rw,nosuid,nodev,noexec,relatime,cpu,cpuacct
Found cgroup option rw,nosuid,nodev,noexec,relatime,cpu,cpuacct, count 8
found blkio in rw,nosuid,nodev,noexec,relatime,blkio
Found cgroup option rw,nosuid,nodev,noexec,relatime,blkio, count 9
found freezer in rw,nosuid,nodev,noexec,relatime,freezer
Found cgroup option rw,nosuid,nodev,noexec,relatime,freezer, count 10
[...]

PR #170, opened 2022-06-10

Hackerl:


Hackerl:
> @Hackerl Thanks for submitting the patch. Can you please add your `Signed-off-by` to the patch, it's a requirement to merge the patch (https://github.com/libcgroup/libcgroup/blob/main/CONTRIBUTING.md#sign-your-work). Also, can you add bug report #168 as the description to the patch, so that we know what the issue is and why is this patch required?

Sorry I didn't pay attention to these details, I will resubmit.

PR #173, opened 2022-06-11

Hackerl:
calling "hasmntopt" to determine if the controller name exists in "mntopt", may cause errors because of "hasmntopt" only match substring.

cpu controller may incorrectly match to cpuset when cpuset mount info appeared before cpu,cpuacct in "/proc/mounts", so we need to validate the last character to make sure the controller name matches exactly.

Signed-off-by: Hackerl <490021209@qq.com>

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1135,11 +1135,17 @@ STATIC int cgroup_process_v1_mnt(char *controllers[], struct mntent *ent,
 	char *strtok_buffer = NULL, *mntopt = NULL;
 	int shared_mnt, duplicate;
 	int i, j, ret = 0;
+	char c = 0;
 
 	for (i = 0; controllers[i] != NULL; i++) {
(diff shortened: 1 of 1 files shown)


Hackerl:
Sorry for my bad English, I can only work with the help of Google Translate.

giani:
hi @Hackerl per the DCO (see https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst ) we need a real name attached to the commit. This is a legal document. Could you please change that way?

Hackerl:
> hi @Hackerl per the DCO (see https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst ) we need a real name attached to the commit. This is a legal document. Could you please change that way?

did you mean to use my real name for "Signed-off-by"?

PR #174, opened 2022-06-11

Hackerl:
calling "hasmntopt" to determine if the controller name exists in "mntopt", may cause errors because of "hasmntopt" only match substring.

cpu controller may incorrectly match to cpuset when cpuset mount info appeared before cpu,cpuacct in "/proc/mounts", so we need to validate the last character to make sure the controller name matches exactly.

Signed-off-by: liupan <490021209@qq.com>


kamalesh-babulal (approved):

drakenclimber:
Sorry for the slow response, @Hackerl.  I caught covid and am only now getting up to speed.  I'll get to this this week.  Thanks!

PR #177, opened 2022-06-27
config: fix string termination issues [v2.0.2]

kamalesh-babulal:
Fix non-terminated string warnings, reported by the Coverity tool:

CID 258293 (#2 of 2): Copy into fixed size buffer (STRING_OVERFLOW).
fixed_size_dest: You might overrun the 32-character fixed-size string
config_namespace_table[namespace_table_index].name by copying name
without checking the length.

fix one another similar string
config_namespace_table[namespace_table_index].mount.path in the same
function cgroup_config_insert_into_namespace_table() by explicitly
terminating by appending '\0';

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -599,14 +599,21 @@ void cgroup_config_cleanup_mount_table(void)
  */
 int cgroup_config_insert_into_namespace_table(char *name, char *nspath)
 {
+	char *ns_tbl_name, *ns_tbl_path;
+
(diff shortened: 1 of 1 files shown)


drakenclimber:
Pushed to main in commit 8eee916573cd9a67713dd645d644d22759f39a69.  Pushed to release-2.0 in commit b54845ab05bdbbf8adc682770fde56c8cb0ce3b7.

PR #178, opened 2022-06-28
Increase the sources column width to 100

kamalesh-babulal:
it is more readable.

This is the first in the series of patch sets.

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -8,8 +8,8 @@
 # We do not follow the negative error numbering, so ignore it.
 --ignore USE_NEGATIVE_ERRNO
 
-# Follow the 80-character limit.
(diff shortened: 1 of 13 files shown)


kamalesh-babulal:
**V2 Changes**
- Increase the `--max-line-length` to 100 in `.checkpatch.conf`
- Don't excess 80 cols for comments

kamalesh-babulal:
** V2 Changes **
- Rebased the patches on top of current origin main `0dd16b472c64`

drakenclimber (approved):

drakenclimber:
Pushed to main.  Thanks, @kamalesh-babulal.

(And stating the obvious here, but this was _not_ pushed to release-2.0.)

PR #179, opened 2022-06-29
Increase the sources column width to 100 [Part 2]

kamalesh-babulal:
it more readable.

This is the second in the series of patch sets.


kamalesh-babulal:
**V2 Changes**

- Rebased the patches on top of the current origin main 0dd16b472c64

drakenclimber (approved):

drakenclimber:
Merged into main.  Thanks, @kamalesh-babulal.

(And, again, this was intentionally not added to release-2.0.)

PR #186, opened 2022-07-07
tools/cgxget: fix a resource leak get_cv_value()

kamalesh-babulal:
Fix a resource leak in get_cv_value(), reported by Coverity tool:

CID 258291 (#1 of 1): Resource leak (RESOURCE_LEAK). leaked_storage:
Variable handle going out of scope leaks the storage it points to.

failure on the realloc(), doesn't free the *handle. Fix it by using the
goto read_end(), that does the job of freeing the handle.

### DIFF
--- a/src/tools/cgxget.c
+++ b/src/tools/cgxget.c
@@ -560,7 +560,7 @@ static int get_cv_value(struct control_value * const cv,
 			tmp = realloc(cv->multiline_value, sizeof(char) *
 				(strlen(cv->multiline_value) + strlen(tmp_line) + 3));
 			if (tmp == NULL)
-				goto end;
+				goto read_end;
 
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 0dd16b472c649936b885838796ab33b351b5695e.  `cgxget` isn't in release-2.0, so this fix is only required in main.

drakenclimber:
@kamalesh-babulal - Do we need a similar fix in cgget?
https://github.com/libcgroup/libcgroup/blob/main/src/tools/cgget.c#L531

kamalesh-babulal:
@drakenclimber Thanks for the review. Good catch, yes we need to fix it in cgget too. Will be sending the patch for it soon.

PR #189, opened 2022-07-11
tools/cgxget: replace strcat() with strncat()

kamalesh-babulal:
Fix copy into fixed size buffer warning, reported by Coverity tool:

CID 258284 (#4 of 4): Copy into fixed size buffer (STRING_OVERFLOW)1.
fixed_size_dest: You might overrun the 4096-character fixed-size string
tmp_val by copying tok without checking the length.

In indent_multiline_value(), warned about the usage of strcat(), that
might overwrite the string. Fix it by replacing strcat() -> strncat()
in the function.

### DIFF
--- a/src/tools/cgxget.c
+++ b/src/tools/cgxget.c
@@ -560,11 +560,11 @@ static int indent_multiline_value(struct control_value * const cv)
 	char *tok, *saveptr = NULL;
 
 	tok = strtok_r(cv->value, "\n", &saveptr);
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
Sorry, looks like the wrong diff. Will force the right diff sooner.

kamalesh-babulal:
**V2 Changes**
- Push the right patch diff

PR #192, opened 2022-07-13
api.c: add ret value fix in cgroup_change_cgroup_flags() [v2.0.2]

kamalesh-babulal:
Fix the unused ret value warning, reported by Coverity:

CID 258304 (#1 of 1): Unused value (UNUSED_VALUE)returned_value:
Assigning value from cgroup_create_template_group(newdest, tmp, flags)
to ret here, but that stored value is overwritten before it can be used.

It turns out that cgroup_change_cgroup_flags(), calls the function
cgroup_create_template_group() to a cgroup based on a template, which
may fail and we don't check the return value for the error.  This fix
adds the check and bails out on error.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -4113,6 +4113,11 @@ int cgroup_change_cgroup_flags(uid_t uid, gid_t gid,
 
 			cgroup_dbg("control group %s is template\n", newdest);
 			ret = cgroup_create_template_group(newdest, tmp, flags);
+			if (ret) {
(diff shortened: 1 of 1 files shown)


drakenclimber:
Pushed to main in commit c46591c5cf0d85b551668d98b33c86a68853b3d1.  Pushed to release-2.0 in commit ae96348967a82fcd18527cb150ac6f7235fb0777

PR #193, opened 2022-07-14
tools/cgsnapshot: add ret value fix in parse_controllers()

kamalesh-babulal:
Fix the unused ret value warning, reported by Coverity:

CID 258275 (#1 of 1): Unused value (UNUSED_VALUE)returned_value:
Assigning value from display_controller_data(controllers, program_name)
to ret here, but that stored value is overwritten before it can be used.

The parse_controllers(), doesn't check for errors in the value returned
by display_controller_data(). The return value might very well contain
an error, that might go unnoticed. Fix it by adding a check for the
return value.


drakenclimber (approved):

drakenclimber:
Pushed to main in commit f2f107c8b3ef0f699c0a3f77aeebbcbeb4f1b8b1.  Pushed to release-2.0 in commit 381834ed04ff4a6e7ad16aa84df5ee6581fae815

PR #198, opened 2022-07-18
api.c: fix string termination in cgroup_get_procname_from_procfs()

kamalesh-babulal:
Fix non-terminated string warning, reported by Coverity tool:

CID 258273 (#2 of 2): String not null terminated (STRING_NULL)6.
string_null: Passing unterminated string buf to strdup, which expects a
null-terminated string.

use snprintf() instead of sprintf(), to string terminate
cgroup_get_procname_from_procfs().

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -5370,7 +5370,7 @@ int cgroup_get_procname_from_procfs(pid_t pid, char **procname)
 
 	/* Get the full patch of process name from /proc/<pid>/exe. */
 	memset(buf, '\0', sizeof(buf));
-	sprintf(path, "/proc/%d/exe", pid);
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
@drakenclimber We have a new kernel update with Ubuntu latest (focal) and it has additional fields in CPU controller files. Can you please pull in https://github.com/libcgroup/libcgroup-tests/pull/59, which describes the new additional fields and fixes all the test cases it relies on?

drakenclimber (approved):

drakenclimber:
Pushed to main in commit 422c0300583a7b450838e4936ba0d51584354831.  Pushed to release-2.0 in commit 9612128edd03ba5085cd0c169685915cd7d6e2c8

PR #200, opened 2022-07-20
api: null terminate readlink buffer in cg_get_procname_from_proc_cmdl…

kamalesh-babulal:
…ine()

Fix readlink buffer null termination warning, reported by Coverity tool:

CID 258276 (#1 of 1): String not null terminated (STRING_NULL).
string_null: Passing unterminated string buf_cwd to strcat, which
expects a null-terminated string.

As per the man pages (man 2 readlink):
"readlink() does not append a null byte to buf.  It  will  (silently)
truncate the contents (to a length of bufsiz characters), in case the
buffer is too small to hold all of the contents."

Explicitly null terminate the buffer passed to readlink() in
cg_get_procname_from_proc_cmdline().

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -5288,6 +5288,9 @@ static int cg_get_procname_from_proc_cmdline(pid_t pid, const char *pname_status
 	if (readlink(path, buf_cwd, sizeof(buf_cwd)) < 0)
 		return ECGROUPNOTEXIST;
 
+	/* readlink doesn't append a null */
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 23bdfd2326a01675a0eac9ab7b601b4a505d219c.  Pushed to release-2.0 in commit bfc9ee791899e57a25b5621205d154a7f91ffff7

PR #203, opened 2022-07-21
api: null terminate readlink buffer in cgroup_get_procname_from_procfs() [v2.0.02]

kamalesh-babulal:
Fix readlink buffer null termination warning, reported by Coverity tool:

CID 258273 (#2 of 2): String not null terminated (STRING_NULL).
string_null: Passing unterminated string buf to strdup, which expects a
null-terminated string.

As per the man pages (man 2 readlink):
"readlink() does not append a null byte to buf.  It  will (silently)
truncate the contents (to a length of bufsiz characters), in case the
buffer is too small to hold all of the contents."

Explicitly null terminate the buffer passed to readlink() in
cgroup_get_procname_from_procfs()

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -5310,6 +5310,9 @@ int cgroup_get_procname_from_procfs(pid_t pid, char **procname)
 		*procname = pname_status;
 		return 0;
 	}
+	/* readlink doesn't append a null */
+	buf[FILENAME_MAX - 1] = '\0';
+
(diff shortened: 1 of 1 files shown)


drakenclimber:
Pushed to main in commit 397554c9df86.  Pushed to release-2.0 in commit 21c50cb860304ad99af920e80e48df941ecf7d75

PR #207, opened 2022-07-24
tools/cgxget: fix the resource leak in convert_cgroup()

kamalesh-babulal:
Fix a resource leak warning reported by the Coverity tool:

CID 258272 (#1-2 of 2): Resource leak (RESOURCE_LEAK). leaked_storage:
Variable cg_converted_list going out of scope leaks the storage it
points to.

while free'ing() the cg_converted_list() via cgroup_free(), wrong array
index variable was passed, causing the resource leak. Fix it by passing
the right index variable.

### DIFF
--- a/src/tools/cgxget.c
+++ b/src/tools/cgxget.c
@@ -769,7 +769,7 @@ out:
 	if (ret != 0 && ret != ECGNOVERSIONCONVERT) {
 		/* The conversion failed */
 		for (j = 0; j < i; j++)
-			cgroup_free(&(cg_converted_list[i]));
+			cgroup_free(&(cg_converted_list[j]));
 	} else {
 		/*
 		 * The conversion succeeded or was unmappable.


kamalesh-babulal:
`cgget` doesn't need this fix, this code is `cgxget` specific.

src/tools/cgxget.c
```diff
@@ -769,7 +769,7 @@ static int convert_cgroups(struct cgroup **cg_list[], int cg_list_len,
 	if (ret != 0 && ret != ECGNOVERSIONCONVERT) {
 		/* The conversion failed */
 		for (j = 0; j < i; j++)
-			cgroup_free(&(cg_converted_list[i]));
```
drakenclimber, line 772: :facepalm: I wrote an embarrassing bug here

drakenclimber (approved):

drakenclimber:
Pushed to main in commit 66799b867c7ace0affa61fba70fb0fab42725499.  cgxget.c isn't in release-2.0, so there's no need to cherry-pick this commit to release-2.0

PR #208, opened 2022-07-25
api.c: check for invalid error code in cgroup_strerror()

kamalesh-babulal:
Fix array overflow warning, reported by the Coverity tool:

CID 258309 (#1 of 1): Out-of-bounds read (OVERRUN). overrun-local:
Overrunning array cgroup_strerror_codes of 32 8-byte elements at element
index 49999 (byte offset 399999) using index code % ECGROUPNOTCOMPILED
(which evaluates to 49999).

there are chances of users passing error codes, resulting in crossing
the upper bound of the cgroup_strerror_codes[], fix it by introducing
bound checks.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -4558,10 +4558,15 @@ cleanup_path:
 
 const char *cgroup_strerror(int code)
 {
+	int idx = code % ECGROUPNOTCOMPILED;
+
 	if (code == ECGOTHER)
 		return strerror_r(cgroup_get_last_errno(), errtext, MAXLEN);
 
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
Looks like Coverity is not still happy, with this patch. Changing the status to draft.

drakenclimber (approved):

drakenclimber:
Pushed to main in commit 37fab4e36b3785698fcc8af14e624cf1e182c183.  Pushed to release-2.0 in commit 728c6f8fb0f760d786156153b74de9010863584c

PR #211, opened 2022-07-26
tools/cgsnapshot: fix out-of-bounds write in parse_controllers() [v2.0.2]

kamalesh-babulal:
Fix Out-of-bounds write warning, reported by Coverity tool:

CID 258289 (#2 of 2): Out-of-bounds write (OVERRUN)16. overrun-local:
Overrunning array controllers of 100 4096-byte elements at element index
100 (byte offset 413695) using index max (which evaluates to 100).

there are chances, that the index variable max dereferences controller
array might be over the array size of 100. Add upper bound checks
to index variable max, so that it doesn't overrun the controller array.


drakenclimber:
Pushed to main in commit de951069e925a697a047c6327a8578ef6b5d2353.  Pushed to release-2.0 in commit 713fa074a0129ba1676673582b2e9c9dc9eb40b1

PR #214, opened 2022-07-28
api.c: check the bytes read in cgroup_register_unchanged_process()

kamalesh-babulal:
Fix ignoring the number of bytes read, warning reported by Coverity
tool:

CID 258288 (#1 of 1): Ignoring number of bytes read (CHECKED_RETURN).
check_return: read(int, void *, size_t) returns the number of bytes
read, but it is ignored.

In cgroup_register_unchanged_process(), the number of byte read/written
using read()/write() are ignored but coverity it warns about the read()
only, let's fix it.


drakenclimber:
Pushed to main in commit 9d266da61515bb47b889a93229302e545da207fe.  Pushed to release-2.0 in commit 67d073b8208f4e64c88777603e671bc8559cd599

PR #215, opened 2022-07-30
daemon/cgrulesengd: check the bytes read in cgre_receive_unix_domain_…

kamalesh-babulal:
…msg()

Fix ignoring the number of bytes read, warning reported by Coverity
tool:

CID 258286 (#2 of 2): Ignoring number of bytes read (CHECKED_RETURN).
check_return: read(int, void *, size_t) returns the number of bytes
read, but it is ignored.

In cgre_receive_unix_domain_msg(), the number of bytes read() is
ignored, while reading from the flag value from the socket. Coverity
warns on not checking the number of bytes read, fix it.

### DIFF
--- a/src/daemon/cgrulesengd.c
+++ b/src/daemon/cgrulesengd.c
@@ -607,6 +607,7 @@ static void cgre_receive_unix_domain_msg(int sk_unix)
 	char path[FILENAME_MAX];
 	struct stat buff_stat;
 	socklen_t caddr_len;
+	size_t ret_len;
 	int fd_client;
 	int flags;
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 23cee33b4d2c856c1c442f7dfed1237a5e52e2a3.  Pushed to release-2.0 in commit 274510c475ae0275ee2daff73de76f1ccee3b9b9.  Thanks, @kamalesh-babulal 

Note that I removed an extra space in the `if` line.

PR #220, opened 2022-08-03
Add man pages for cgxget and cgxset

drakenclimber:


kamalesh-babulal (approved):
@drakenclimber Thanks for working on adding the man pages.

PR #222, opened 2022-08-05
api.c: add precision to fscanf(), in cgroup_get_all_controller_next()

kamalesh-babulal:

CID 258300 (#1 of 1): Calling risky function

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -5537,7 +5537,11 @@ int cgroup_get_all_controller_next(void **handle, struct controller_data *info)
 	if (!info)
 		return ECGINVAL;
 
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 4036f4fe4334572a4669dfece01ff24dbdd56477.  Pushed to release-2.0 in commit 97aa5a67f9ed067a281cb77a06546ab09191064c.  Pushed to release-3.0 in commit aad6b721131737bc4ac872793dbf42e0b0ae7fde

PR #225, opened 2022-08-07
api.c: add precision to fscanf(), in cgroup_populate_controllers()

kamalesh-babulal:

CID 258305 (#1 of 1): Calling risky function


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 04169a9bef0ee77c18cd5a2e2d9f6e49f6057e32.  Pushed to release-2.0 in commit 5ea19ac2faa16c083c769ad068beee0e4cdc0aad.  Pushed to release-3.0 in commit 98ef774eaae6e2279caaa1aa8582350e7594377c

PR #227, opened 2022-08-09
api.c: add precision to fscanf(), in cgroup_get_current_controller_pa…

kamalesh-babulal:
…th()

CID 258301 (#1 of 1): Calling risky function

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -4526,7 +4526,11 @@ int cgroup_get_current_controller_path(pid_t pid, const char *controller, char *
 		char *token;
 		int num;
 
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit ac05eb49c498aa5047070a4620456fa7a10e1910.  Pushed to release-2.0 in commit a56103d2f833e3a1ba532d2a8b0b222fd1cd1bfd.  Pushed to release-3.0 in commit b897bb8b0b2ccb200f245c10b6a48d97189200fa.

Thanks, @kamalesh-babulal

PR #231, opened 2022-08-11
api.c: fix TOCTOU in cg_chmod_path()

kamalesh-babulal:
Fix TOCTOU warning, reported by Coverity Tool:

CID 258267 (#1 of 1): Time of check time of use (TOCTOU)2.
fs_check_call: Calling function stat to perform check on path.

in cg_chmod_path(), the file name is stat() and not immediately followed
by the file operation. One way to fix it, open the file and use the file
descriptor to manipulate the file.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -212,16 +212,20 @@ int cg_chmod_path(const char *path, mode_t mode, int owner_is_umask)
 {
 	mode_t mask = -1U;
 	struct stat buf;
+	int fd = -1;
 
 	if (owner_is_umask) {
 		mode_t umask, gmask, omask;
 
+		fd = open(path, O_RDONLY);
+		if (fd == -1)
+			goto fail;
 		/*
 		 * Use owner permissions as an umask for group and others
 		 * permissions because we trust kernel to initialize owner
 		 * permissions to something useful.  Keep SUID and SGID bits.
 		 */
-		if (stat(path, &buf) == -1)
+		if (fstat(fd, &buf) == -1)
 			goto fail;
 
 		/* 0700 == S_IRWXU */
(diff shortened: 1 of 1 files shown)


src/api.c
```diff
@@ -232,7 +236,7 @@ int cg_chmod_path(const char *path, mode_t mode, int owner_is_umask)
 		mask = umask|gmask|omask|S_ISUID|S_ISGID|S_ISVTX;
 	}
 
-	if (chmod(path, mode & mask))
+	if (fchmod(fd, mode & mask))
 		goto fail;
 
 	return 0;
```
drakenclimber, line 242: Don't you need to free the `fd` prior to this return as well?

kamalesh-babulal: @drakenclimber Thanks for the review. Yes, it needs to closed.

kamalesh-babulal:
**v2 Changes:**
- close `fd` before returning (found by Tom Hromatka)
- `open()` return int `fd` not a `FILE`, so convert the check from `if (fd)` to `if (fd > 0)`.

drakenclimber (approved):

drakenclimber:
Pushed to main in commit 8b9665c29cb812c255687d6e5cb5262a6fe30a40.  Pushed to release-3.0 in commit c812c8f15171ef182fdb1b0b92478bf5907b4e03.  Pushed to release-2.0 in commit 96db65fbb5296d4e66d1eb444c70d29b556e7eef

drakenclimber:
@kamalesh-babulal.  I just merged this, but I think we still have a problem.  What if `owner_is_umask` is false?  Then the `fd` doesn't get opened but we still try to `fchmod()` and `close()` it later?

kamalesh-babulal:
@drakenclimber Thanks for the review. Good catch, the patch is not complete I have opened #240 to fix the remaining unhandled case with this commit. Can we please close this PR and have the discussion there instead?

PR #232, opened 2022-08-11
api.c: fix TOCTOU in cg_chmod_path()

kamalesh-babulal:
Fix TOCTOU warning, reported by Coverity Tool:

CID 258267 (#1 of 1): Time of check time of use (TOCTOU)2.
fs_check_call: Calling function stat to perform check on path.

in cg_chmod_path(), the file name is stat() and not immediately followed
by the file operation. One way to fix it, open the file and use the file
descriptor to manipulate the file.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -205,17 +205,21 @@ int cg_chmod_path(const char *path, mode_t mode, int owner_is_umask)
 {
 	struct stat buf;
 	mode_t mask = -1U;
+	int fd;
 
 	if (owner_is_umask) {
 		mode_t umask, gmask, omask;
 
+		fd = open(path, O_RDONLY);
+		if (fd == -1)
+			goto fail;
 		/*
 		 * Use owner permissions as an umask for group and others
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
**v2 Changes:**
- `open()` return int `fd` not a `FILE`, so convert the check from `if (fd)` to `if (fd > 0)`.

drakenclimber:
Pushed to main in commit 8b9665c29cb812c255687d6e5cb5262a6fe30a40.  Pushed to release-3.0 in commit c812c8f15171ef182fdb1b0b92478bf5907b4e03.  Pushed to release-2.0 in commit 96db65fbb5296d4e66d1eb444c70d29b556e7eef

But see this [comment](https://github.com/libcgroup/libcgroup/pull/231#issuecomment-1226109149); I don't think we're quite done yet with this issue.

kamalesh-babulal:
I have opened #241 to fix the remaining unhandled case with this commit. Can we please close this PR and have the discussion there instead?

PR #233, opened 2022-08-12
api.c: fix cgroup_attach_task_pid() behavior

kamalesh-babulal:
Originally `cgroup_attach_task_pid(cgroup, tid)`, allowed the passing of
struct cgroup *cgroup or NULL as a valid argument. In the case of cgroup
being passed, the tid was attached to the cgroup by writing into the
'tasks' file of the cgroup.  It gets interesting when the NULL is
passed, the idea is to migrate the task to the cgroup_root of the
hierarchy in which the task is currently running in.

Consider a simple legacy cgroup controller hierarchy:

        cpu,cpuacct             memory          cpuset
          /    \                /               /
         C1    C2              M1              U1
         |                     |             /   \
         t1                    t1          UC1  UC2
                                                 |
						     t1
passing NULL:

        cpu,cpuacct             memory          cpuset
        /    \    |             /    |          /    |
        C1    C2  t1            M1   t1        U1    t1
                                             /   \
					        UC1  UC2

Depending upon the cgroup setup, legacy (only cgroup v1), unified
(cgroup v2) or hybrid (controllers mounted in the cgroup v1 and some
of the functionalities available from cgroup v2).  This behavior got
changed in the due course and will cause a segfault, when passed with
NULL as the cgroup.

Reproducer:
```
int main(int argc, char *argv[])
{
        pid_t pid;
        int ret;

        ret = cgroup_init();
        if (ret) {
                printf("cgroup_init failed with %s\n",
cgroup_strerror(ret));
                exit(1);
        }

        pid = atoi(argv[1]);
        ret = cgroup_attach_task_pid(NULL, pid);
        if (ret)
                printf("attaching %d to cgroup %s failed\n",
cgroup_strerror(ret));

        return ret;
}
```
Re-introduce the original behavior, that considers all the three
possible cgroup modes and performances of the migration based on the
modes. `fscanf()` is in-efficient in parsing `/proc/<pid>/cgroup` fields
with empty values, this is handled by a helper function
`proc_pid_cgroup_parser()`.

It was discovered by the Coverity tool as Dereference after null check:

CID 258278 (#1 of 1): Dereference after null check (FORWARD_NULL).
var_deref_op: Dereferencing null pointer cgroup.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -2142,11 +2142,101 @@ static int cgroup_build_tid_path(const char * const ctrl_name, char *path)
 	return ret;
 }
 
-static int cgroup_attach_task_tid(struct cgroup *cgrp, pid_t tid, bool move_tids)
+/**
+ * proc_pid_cgroup_parser() - Reads a line from /proc/<pid>/cgroup and parses it
+ * @fp The /proc/<pid>/cgroup, file handle
+ * @num Hierarchy number is filled from parsed line
+ * @controllers Controllers names are filled from the parsed line
+ * @cgrp_path Cgroup path is filled from the parsed line
+ *
+ * this function read the next line from the file handle and parses it.
+ * On success, returns 0 and fills the values into the function arguments passed and
+ * failure returns an error and the function arguments might contain stale values.
+ */
+static int proc_pid_cgroup_parser(FILE **fp, int *num, char *controllers, char *cgrp_path)
 {
+	/*
+	 * /proc/<pid>/cgroup:
+	 * 7:memory:/user.slice/user-1000.slice/session-3.scope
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
The comments and code need future polishing.  This patch is not ready for merging yet.

kamalesh-babulal:
**V2 Changes:**
- Rebased to current main.

kamalesh-babulal:
**v3 Changes**:
- Rebased to current main

kamalesh-babulal:
This issue has already been fixed by https://github.com/libcgroup/libcgroup/pull/275

PR #235, opened 2022-08-23
log: introduce CGROUP_LOG_CONT level for logging 

kamalesh-babulal:
There are cases, where we might want to print a very long/multiline log
message to the user. We could call the cgroup_log(), multiple times to
fit the log message, but the downside is that every time the
cgroup_log() called, the log level is prefixed to the message, hence
introducing loglevel char string in the mid of the log message.
For example, calling the cgroup_warn() twice to print a long warning:
```
cgroup_warn("cgroup v1 expects /proc/cgroup, check if ");
cgroup_warn("/proc mounted with subset=pid option\n");
```
output:
-------
```
$ sudo CGROUP_LOGLEVEL=DEBUG ./src/tools/cgget -g cpu:user.slice
...
Warning: cgroup v1 expects /proc/cgroup, check if Warning: /proc mounted
with subset=pid option
```
Introduce a new logging level, CGROUP_LOG_CONT and cgroup_cont() macro,
that will continue printing the log message, when loglevel is set to other
than default log level. The above code can be rewritten as:
```
cgroup_warn("cgroup v1 expects/proc/cgroup, check if ");
cgroup_cont("/proc mounted with subset=pid option\n");
```
output:
-------
```
$ sudo CGROUP_LOGLEVEL=DEBUG ./src/tools/cgget -g cpu:user.slice
...
Warning: cgroup v1 expects /proc/cgroup, check if /proc mounted with
subset=pid option
```
Also, make comsetic changes to cgroup_*().

### DIFF
--- a/include/libcgroup/log.h
+++ b/include/libcgroup/log.h
@@ -72,10 +72,15 @@ extern "C" {
  */
 enum cgroup_log_level {
 	/**
+	 * Continue printing the log message, with the previous log level.
+	 * Used to print log messages without the line break.
+	 */
+	CGROUP_LOG_CONT = 0,
+	/**
 	 * Something serious happened and libcgroup failed to perform requested
 	 * operation.
 	 */
-	CGROUP_LOG_ERROR = 1,
+	CGROUP_LOG_ERROR,
 	/**
 	 * Something bad happened but libcgroup recovered from the error.
 	 */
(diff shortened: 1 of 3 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commits 24b5087cbf80 and db2039f6ad75.  Pushed to release-3.0 in commits 727b6ba53afe and 3966b27e44c0.  I'm not backporting this to release-2.0

Thanks, @kamalesh-babulal

PR #236, opened 2022-08-24
api: Use GNU strerror_r when available

kraj:
GNU strerror_r is only available in glibc, musl impelents the XSI
version which is slightly different, therefore check if GNU version is
available before using it, otherwise use the XSI compliant version.

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -183,6 +183,11 @@ AC_FUNC_REALLOC
 AC_FUNC_STAT
 AC_CHECK_FUNCS([getmntent hasmntopt memset mkdir rmdir strdup])
 
+orig_CFLAGS="$CFLAGS"
+CFLAGS="$CFLAGS -D_GNU_SOURCE"
+AC_FUNC_STRERROR_R
+CFLAGS="$orig_CFLAGS"
+
 AC_SEARCH_LIBS(
 	[fts_open],
 	[fts],
--- a/src/api.c
+++ b/src/api.c
@@ -4571,9 +4571,13 @@ const char *cgroup_strerror(int code)
 {
 	int idx = code % ECGROUPNOTCOMPILED;
 
-	if (code == ECGOTHER)
+	if (code == ECGOTHER) {
+#ifdef STRERROR_R_CHAR_P
 		return strerror_r(cgroup_get_last_errno(), errtext, MAXLEN);
-
+#else
+		return strerror_r(cgroup_get_last_errno(), errtext, sizeof (errtext)) ? "unknown error" : errtext;
+#endif
+	}
 	if (idx >= sizeof(cgroup_strerror_codes)/sizeof(cgroup_strerror_codes[0]))
 		return "Invalid error code";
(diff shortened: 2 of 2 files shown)


kamalesh-babulal:
Hi  @kraj,

Thanks for reporting and fixing the warning on using `strerror_r()` with musl.
I tried reproducing using the version `musl-libc-1.2.3-1`, `musl-gcc-1.2.3-1` 
shipped with my `Fedora 35` VM, using following commands to compile the
libcgroup:
```
$ CC="musl-gcc" ./bootstrap.sh
...
...
configure: error: Unable to find the fts_open() function
```
It fails at various places, during configure. Any suggestions on how to resolve
them? I did a couple of hacks to bypass the configure step and as expected it failed
at make, among the various warnings/errors, your patch fixes with the following warning:
```
api.c: In function ‘cgroup_strerror’:
api.c:4643:24: warning: returning ‘int’ from a function with return type ‘const char *’ makes pointer from integer without a cast [-Wint-conversion]
 4643 |                 return strerror_r(cgroup_get_last_errno(), errtext, MAXLEN);
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
after the patch, the configure prints the availability of `strerror_r`:
```
...
checking whether strerror_r is declared... yes
checking for strerror_r... yes
checking whether strerror_r returns char *... no
...
```
we might need some work to support musl.

kraj:
> 

you need `fts` devel installed on system see
https://git.yoctoproject.org/poky/tree/meta/recipes-core/libcgroup/libcgroup_3.0.0.bb#n19

and you can call build with  `make LIBS=-lfts`
https://git.yoctoproject.org/poky/tree/meta/recipes-core/libcgroup/libcgroup_3.0.0.bb#n20

kamalesh-babulal:
Sorry for the delay in the response, I was on a vacation. Thanks for the link, I will look into it after Linux Plumbers Conference.

thesamesam:
@kamalesh-babulal Did you have a chance to look again? Thanks!

kamalesh-babulal:
@thesamesam My sincere apologies, will look into it this week. Again, sorry for the delay.

thesamesam:
No worries at all! Thanks for the fast response!

kamalesh-babulal:
@kraj @thesamesam Thank you for your contribution and help. Sorry for getting to it so late.

drakenclimber (approved):

drakenclimber:
Thanks, @kraj, @thesamesam, and @kamalesh-babulal.  Pushed to main in commit 942ef655237b.  Pushed to release-3.0 in commit 1211e1621d3d

drakenclimber:
Note this will not be backported to the release-2.0 branch as that branch is accepting only mission-critical bug fixes and is slated to be retired after its next release.

PR #237, opened 2022-08-24
support /proc mounted with subset=pid [v2.0]

kamalesh-babulal:
`/proc` filesystem can be mounted with `subset=pid` as one of its mount
options.  This option hides all the top-level files and directories,
those are not related to processes. The cgroup v1 filesystem depends
on the `/proc/cgroups` to populate the cgroups controllers and will fail
during the `cgroup_init()` phase, when not available, whereas cgroup v2
considers this as a deprecated file and recommends reading the list of
controller from `<unified mount point>/cgroup.controllers`[1].

This patch series contains two patches.  The first patch replaces the
`/proc/mounts` with `/proc/mounts` in `src/api.c`,  as the file to read the
current mount points in the system and the second patch adds the support
for reading the controllers from the `cgroup.controller`, when the  `/proc`
mount is mounted with `subset=pid` option only when the system is booted
with the unified mode and will fail to initialize in the case we find the
cgroup v1 mounted, i.e, the system booted with legacy or hybrid mode.
   
[1] https://docs.kernel.org/admin-guide/cgroup-v2.html#deprecated-v1-core-features

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1299,6 +1299,75 @@ out:
 	return ret;
 }
 
+/*
+ * Parses the mount options of the given mount point and checks for the
+ * option in the list of mount options.
+ * @mnt: Mount point name to search for mount points.
+ * @mnt_opt: Mount option to be searched.
+ * @is_set: Set to 1, when mount option is found, 0 otherwise.
+ *
+ * Returns 0, in case of success and ECGOTHER on failure.
+ */
+static int check_mount_point_opt(const char *mnt, const char *mnt_opt,
+				 int * const is_set)
+{
+	struct mntent *ent, *temp_ent = NULL;
+	char mntent_buffer[4 * FILENAME_MAX];
+	char *mntopt = NULL;
+	char mnt_opt_delim;
+	FILE *proc_mount;
+	int ret = 0;
+
+	if (!mnt || !mnt_opt || !is_set)
+		return ECGINVAL;
+
+	proc_mount = setmntent(mnt, "r");
+	if (!proc_mount) {
+		cgroup_err("Error: cannot open %s:%s ", mnt, strerror(errno));
+		last_errno = errno;
+		ret = ECGOTHER;
+		goto err;
+	}
+
+	temp_ent = (struct mntent *) malloc(sizeof(struct mntent));
+	if (!temp_ent) {
+		last_errno = errno;
+		ret = ECGOTHER;
+		goto err;
+	}
+
+	ent = getmntent_r(proc_mount, temp_ent, mntent_buffer,
+			  sizeof(mntent_buffer));
+	if (!ent) {
+		last_errno = errno;
+		ret = ECGOTHER;
+		goto err;
+	}
+
+	*is_set = 0;
+	while ((mntopt = hasmntopt(ent, mnt_opt))) {
+		mnt_opt_delim = mntopt[strlen(mnt_opt)];
+		if (mnt_opt_delim == '\0' || mnt_opt_delim == ',') {
+			*is_set = 1;
+			break;
+		}
+	}
+
+	if (*is_set == 0) {
+		cgroup_dbg("Debug: %s, not found in the list of ", mnt_opt);
+		cgroup_dbg("mount options of %s\n", mnt);
+	}
+
+err:
+	if (proc_mount)
+		endmntent(proc_mount);
+
+	if (temp_ent)
+		free(temp_ent);
+
+	return ret;
+}
+
 /**
  * cgroup_init(), initializes the MOUNT_POINT.
  *
@@ -1322,6 +1391,8 @@ int cgroup_init(void)
 	int err;
 	char *buf = NULL;
 	char mntent_buffer[4 * FILENAME_MAX];
+	char mnt_opt[] = "subset=pid";
+	int mnt_opt_set;
 
 	cgroup_set_default_logger(-1);
 
@@ -1339,13 +1410,25 @@ int cgroup_init(void)
 	memset(&cg_mount_table, 0, sizeof(cg_mount_table));
 
 	proc_cgroup = fopen("/proc/cgroups", "re");
-
 	if (!proc_cgroup) {
-		cgroup_err("Error: cannot open /proc/cgroups: %s\n",
-				strerror(errno));
-		last_errno = errno;
-		ret = ECGOTHER;
-		goto unlock_exit;
+		cgroup_warn("Warning: cannot open /proc/cgroups: %s\n",
+			    strerror(errno));
+		ret = check_mount_point_opt("/proc/self/mounts", mnt_opt,
+					    &mnt_opt_set);
+		if (ret)
+			goto unlock_exit;
+
+		if (!mnt_opt_set) {
+			ret = ECGINVAL;
+			goto unlock_exit;
+		}
+
+		/*
+		 * /proc, mounted with subset=pid is valid. cgroup v2 doesn't
+		 * depend on /proc/cgroups to parse the available controllers.
+		 */
+		ret = 0;
+		goto cgroup_unified;
 	}
 
 	/*
@@ -1385,6 +1468,7 @@ int cgroup_init(void)
 	}
 	controllers[i] = NULL;
 
+cgroup_unified:
 	proc_mount = fopen("/proc/self/mounts", "re");
 	if (proc_mount == NULL) {
 		cgroup_err("Error: cannot open /proc/self/mounts: %s\n",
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
Backported the review comments from @drakenclimber on the patch posted for `v3.0+` release at #238

0n-s:
Ehm, excuse me for thinking this was an older version of the patchset & not a backport until yesterday. （─.─||）

> @0n-s Can you check if the patch helps

I've been testing random `cgget` commands & `pam_cgroup` on a cgroup2-only system (`cgroup_no_v1=all,named`) for about a day now. It's been working well.

Most recently I tried a patch that removes `/proc/cgroups` in the kernel entirely, regardless of `subset=pid`.
```
--- a/kernel/cgroup/cgroup.c
+++ b/kernel/cgroup/cgroup.c
@@ -6000,7 +6000,6 @@ int __init cgroup_init(void)
 	WARN_ON(sysfs_create_mount_point(fs_kobj, "cgroup"));
 	WARN_ON(register_filesystem(&cgroup_fs_type));
 	WARN_ON(register_filesystem(&cgroup2_fs_type));
-	WARN_ON(!proc_create_single("cgroups", 0, NULL, proc_cgroupstats_show));
 #ifdef CONFIG_CPUSETS
 	WARN_ON(register_filesystem(&cpuset_fs_type));
 #endif
```
(applied against commit `10d4879f9ef01cc6190fafe4257d06f375bab92c` of linux.git)
This was essentially purely to my sate my curiosity, but it didn't work with this patch (this patch to the kernel & this patch to libcgroup) applied. OFC, this isn't really any matter in practice & it'd be entirely reasonable to ignore this, but it would be nice if the patch just checked the lack of `/proc/cgroups` & fell back to cgroup2 instead of checking mount options on procfs. `subset=pid` is a procfs feature that got introduced retroactively, & I can imagine a theoretical knob to compile out the cgroup1 code removing that too. Not that I know of anything like that being in the works, but doing it like that would reduce the need for another workaround in libcgroup should that come. Just an idea, though. FTTB your patch as it is will fix the issue on all current kernels in any configuration as far as I can tell, so it does the job in practice.

Without the aforementioned kernel patch, though, everything has been working as expected & basically matching the behavior when `/proc/cgroups` is present.

> also provide your name and email

As mentioned before, my GitHub username should be fine. But if you insist on the specific format, then use this: `Reported-by: ns <0n-s@users.noreply.github.com>`

kamalesh-babulal:
> Ehm, excuse me for thinking this was an older version of the patchset & not a backport until yesterday

@0n-s, as you would have noticed, there are two release branches, one tagged as `v2.0+` and another tagged as `v3.0+`. Both of
them diverge a bit from each other, hence rebase of a patch written against one version could not apply to the other, without some porting. I generally suffix the pull request title with `[v2.0]` for patches that apply cleanly on the v2.0 code base.

> I've been testing random cgget commands & pam_cgroup on a cgroup2-only system (cgroup_no_v1=all,named) for about a day now. It's been working well.

Thank you for testing it extensively, please do let us know your likes/comments/improvements you would like to see or a patch with PR would be great.

> This was essentially purely to my sate my curiosity, but it didn't work with this patch (this patch to the kernel & this patch to libcgroup) applied. OFC, this isn't really any matter in practice & it'd be entirely reasonable to ignore this, but it would be nice if the patch just checked the lack of /proc/cgroups & fell back to cgroup2 instead of checking mount options on procfs. subset=pid is a procfs feature that got introduced retroactively, & I can imagine a theoretical knob to compile out the cgroup1 code removing that too. Not that I know of anything like that being in the works, but doing it like that would reduce the need for another workaround in libcgroup should that come. Just an idea, though. FTTB your patch as it is will fix the issue on all current kernels in any configuration as far as I can tell, so it does the job in practice.

I completely agree, that if the `/proc/cgroups` is not created by the Linux Kernel the check for subset=pid can be removed and should fall back to reading the controllers from `<cgroup v2>/cgroup.controllers` instead, like in the case of the above kernel patch.

libcgroup is tightly coupled with Linux Kernel and another side of the argument is that the `/proc/cgroups` is supposed to be created during Kernel booting in `cgroup_init()` or else a `WARN_ON()` could have been trigger on failure to create the file, but the Kernel could continue boot. In this case, the libcgroup tools or any application using libcgroup would fail, because `cgroup_init()` expects `/proc/cgroup` and libcgroup would complain about missing `/proc/cgroup`.

The only scenario, for now, where it's valid for `/proc/cgroup` not to be present is when the user mounts `/proc/cgroup` with `subset=pid` as one of its mount points. The check for mount option is needed to support the valid combinations in Linux Kernel and yes, the day we have a knob or `/proc/cgroup` gets removed, we should revisit and fix the libcgroup behavior too.

> As mentioned before, my GitHub username should be fine. But if you insist on the specific format, then use this: Reported-by: ns <0n-s@users.noreply.github.com>

Thanks, I was not sure if you would read both the PR, so had added it in both. I have pushed the patches with your GitHub username.

0n-s:
> @0n-s, as you would have noticed, there are two release branches…

Yeah, I got that. Just guessed wrong a few days ago, so there was a bit of a delay between me knowing about these patches & testing them.

> Thank you for testing it extensively, please do let us know your likes/comments/improvements you would like to see or a patch with PR would be great.

I dunno, it fixes the issue in the bug report without anything new showing up (AFAIK). As far as I'm concerned, it's perfect.


> libcgroup is tightly coupled with Linux Kernel and another side of the argument is that the `/proc/cgroups` is supposed to be created during Kernel booting in `cgroup_init()` or else a `WARN_ON()` could have been trigger on failure to create the file, but the Kernel could continue boot. In this case, the libcgroup tools or any application using libcgroup would fail, because `cgroup_init()` expects `/proc/cgroup` and libcgroup would complain about missing `/proc/cgroup`.
> 
> The only scenario, for now, where it's valid for `/proc/cgroup` not to be present is when the user mounts `/proc/cgroup` with `subset=pid` as one of its mount points. The check for mount option is needed to support the valid combinations in Linux Kernel and yes, the day we have a knob or `/proc/cgroup` gets removed, we should revisit and fix the libcgroup behavior too.

Sure, it was mainly a suggestion so if such a config knob came through (the idea's quite plausible to me since cgroup1 & cgroup2 conflict with each other to a non-negligible extent when used simultaneously, hence `cgroup_no_v1=`), it would just work without any extra patches.

Also, I would advise anyone _not_ to consider `WARN()` as a valid condition. `WARN()` is used for "impossible" conditions, which means a kernel bug. & working against those is really just an exercise in futility. :)
LKML has already worded this better than I can here, particularly these bits:
> What WARN means is that an unexpected condition occurred which means the kernel itself is in an unknown state.
> …
> Your only policy choices when hitting WARN are
>
>   1. Accept the risk and continue degraded operation, or
>   2. reset the system to a known good state [by rebooting].

([source](https://lwn.net/ml/linux-kernel/88093da62a4b85f015423cbd1ec2f5ad6eb0c5da.camel@HansenPartnership.com))

& believe me, the "if it boots, it's valid" argument can be taken to some absolutely broken conclusions even without reaching anywhere near its logical extreme (think executables failing to start for no reason, filesystems randomly vanishing, processes hanging indefinitely, random process memory corruption, & more as part of not particularly uncommon effects of a dmesg-worthy kernel bug). Most distro kernels I'm familiar with don't even panic on oops by default…

I could go on & on about that with increasingly nightmarish examples, but the gist is, userspace aiming to work past a kernel assert is a lost cause. So it's not very wise (quite the opposite, really).

> Thanks, I was not sure if you would read both the PR, so had added it in both. I have pushed the patches with your GitHub username.

Thank you!

drakenclimber (approved):

drakenclimber:
Pushed to main in commits 16be308b037b and c83cb5a2d2ff.  Pushed to release-2.0 in commits 1686eb6f24b3 and d7cd75e24c27.  Pushed to release-3.0 in commits 860ac4456c6a and 050dd1c0abb1.

Thanks, @0n-s and @kamalesh-babulal

kamalesh-babulal:
> Yeah, I got that. Just guessed wrong a few days ago, so there was a bit of a delay between me knowing about these patches & testing them.
> I dunno, it fixes the issue in the bug report without anything new showing up (AFAIK). As far as I'm concerned, it's perfect.

Sorry for the delay in the response, I was on a vacation. Thanks again for testing.

> Sure, it was mainly a suggestion so if such a config knob came through (the idea's quite plausible to me since cgroup1 & cgroup2 conflict with each other to a non-negligible extent when used simultaneously, hence cgroup_no_v1=), it would just work without any extra patches.

Agreed, it would be seamless and the check will be not required for `subset=pid`

> Also, I would advise anyone not to consider WARN() as a valid condition. WARN() is used for "impossible" conditions, which means a kernel bug. & working against those is really just an exercise in futility. :)
LKML has already worded this better than I can here, particularly these bits:
>  What WARN means is that an unexpected condition occurred which means the kernel itself is in an unknown state.
    …
    Your only policy choices when hitting WARN are

        Accept the risk and continue degraded operation, or
        reset the system to a known good state [by rebooting].

([source](https://lwn.net/ml/linux-kernel/88093da62a4b85f015423cbd1ec2f5ad6eb0c5da.camel@HansenPartnership.com))

> & believe me, the "if it boots, it's valid" argument can be taken to some absolutely broken conclusions even without reaching anywhere near its logical extreme (think executables failing to start for no reason, filesystems randomly vanishing, processes hanging indefinitely, random process memory corruption, & more as part of not particularly uncommon effects of a dmesg-worthy kernel bug). Most distro kernels I'm familiar with don't even panic on oops by default…

Yes, that's true, thanks for the link.  `if it boots, it's valid` sums it up all, I was referring to the `WARN()` as a debugging step to check on why `cgroup_init()` failed, even with `/proc/` not mounted with `subset=pid`

> I could go on & on about that with increasingly nightmarish examples, but the gist is, userspace aiming to work past a kernel assert is a lost cause. So it's not very wise (quite the opposite, really).

:)

PR #238, opened 2022-08-24
support /proc mounted with subset=pid [v3.0+] 

kamalesh-babulal:
`/proc` filesystem can be mounted with `subset=pid` as one of its mount
options. This option hides all the top-level files and directories,
those are not related to processes. The cgroup v1 filesystem depends
on the `/proc/cgroups` to populate the cgroups controllers and will fail
during the `cgroup_init()` phase, when not available, whereas cgroup v2
considers this as a deprecated file and recommends reading the list of
controller from `<unified mount point>/cgroup.controllers`[1].

This patch series contains two patches. The first patch replaces the
`/proc/mounts` with `/proc/mounts` in `src/api.c`, as the file to read the
current mount points in the system and the second patch adds the support
for reading the controllers from the cgroup.controller, when the /proc
mount is mounted with `subset=pid` option only when the system is booted
with the unified mode and will fail to initialize in the case we find the
cgroup v1 mounted, i.e, the system booted with legacy or hybrid mode.

[1] https://docs.kernel.org/admin-guide/cgroup-v2.html#deprecated-v1-core-features

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1407,7 +1407,7 @@ err:
 }
 
 /*
- * Reads /proc/mounts and populates the cgroup v1/v2 mount points into the
+ * Reads /proc/self/mounts and populates the cgroup v1/v2 mount points into the
  * global cg_mount_table.
  * This function should be called with cg_mount_table_lock taken.
  */
@@ -1419,9 +1419,9 @@ static int cgroup_populate_mount_points(char *controllers[CG_CONTROLLER_MAX])
 	FILE *proc_mount;
 	int ret = 0;
 
-	proc_mount = fopen("/proc/mounts", "re");
+	proc_mount = fopen("/proc/self/mounts", "re");
 	if (proc_mount == NULL) {
-		cgroup_err("cannot open /proc/mounts: %s\n", strerror(errno));
+		cgroup_err("cannot open /proc/self/mounts: %s\n", strerror(errno));
 		last_errno = errno;
 		ret = ECGOTHER;
 		goto err;
@@ -1537,7 +1537,7 @@ static int cg_test_mounted_fs(void)
 	FILE *proc_mount = NULL;
 	int ret = 1;
 
-	proc_mount = fopen("/proc/mounts", "re");
+	proc_mount = fopen("/proc/self/mounts", "re");
 	if (proc_mount == NULL)
 		return 0;
 
--- a/src/api.c
+++ b/src/api.c
@@ -1331,6 +1331,73 @@ static void cgroup_free_cg_mount_table(void)
 }
 
 /*
+ * Parses the mount options of the given mount point and checks for the
+ * option in the list of mount options and sets is_set accordingly.
+ * @mnt: Mount point name to search for mount points.
+ * @mnt_opt: Mount option to be searched.
+ * @is_set: Set to 1, when mount option is found, 0 otherwise.
+ *
+ * Returns 0, in case of success and ECGOTHER on failure.
+ */
+static int check_mount_point_opt(const char *mnt, const char *mnt_opt, int * const is_set)
+{
+	struct mntent *ent, *temp_ent = NULL;
+	char mntent_buffer[4 * FILENAME_MAX];
+	char *mntopt = NULL;
+	char mnt_opt_delim;
+	FILE *proc_mount;
+	int ret = 0;
+
+	if (!mnt || !mnt_opt || !is_set)
+		return ECGINVAL;
+
+	proc_mount = setmntent(mnt, "r");
+	if (!proc_mount) {
+		cgroup_err("cannot open %s: %s\n", mnt, strerror(errno));
+		last_errno = errno;
+		ret = ECGOTHER;
+		goto err;
+	}
+
+	temp_ent = (struct mntent *) malloc(sizeof(struct mntent));
+	if (!temp_ent) {
+		last_errno = errno;
+		ret = ECGOTHER;
+		goto err;
+	}
+
+	ent = getmntent_r(proc_mount, temp_ent,	mntent_buffer, sizeof(mntent_buffer));
+	if (!ent) {
+		last_errno = errno;
+		ret = ECGOTHER;
+		goto err;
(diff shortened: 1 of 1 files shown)


src/api.c
```diff
@@ -1321,6 +1321,60 @@ static void cgroup_free_cg_mount_table(void)
...
+/*
+ * Parses the mount options of the given mount point and checks for the
+ * option in the list of mount options.
+ * @mnt: Mount point name to search for mount points.
+ * @mnt_option: Mount option to be searched.
+ *
+ * Returns 0, in case of success and ECGOTHER on failure.
+ */
+static int check_mount_point_opt(char *mnt, char *mnt_option)
+{
```
drakenclimber, line 1333: It feels unintuitive to return `0` for "yes, this opt is set" and an error code for all other cases.  If it's not set, the user is still getting an error; should the user abort or does the error simply mean the opt isn't set?

Would it make more sense to pass in an `is_set` parameter that this function could write to?
```
static int check_mount_point_opt(char *mnt, char *mnt_option, int * const is_set)
```

src/api.c
```diff
@@ -1321,6 +1321,60 @@ static void cgroup_free_cg_mount_table(void)
...
+	mntopt = hasmntopt(ent, mnt_option);
+	if (!mntopt) {
+		cgroup_dbg("%s, not found in the list of ", mnt_option);
+		cgroup_cont("mount options of %s\n", mnt);
+		ret = ECGOTHER;
+	}
+
+err:
+	if (proc_mount)
+		fclose(proc_mount);
```
drakenclimber, line 1371: Per `setmntent`'s man page, should we be using `endmntent()` rather than `fclose()`?

```
       The setmntent() function opens the filesystem description file filename and returns a file pointer which can
       be  used  by  getmntent().  The argument type is the type of access required and can take the same values as
       the mode argument of fopen(3).   The  returned  stream  should  be  closed  using  endmntent()  rather  than
       fclose(3).
```

src/api.c
```diff
@@ -1321,6 +1321,60 @@ static void cgroup_free_cg_mount_table(void)
...
+	}
+
+	ent = getmntent_r(proc_mount, temp_ent,	mntent_buffer, sizeof(mntent_buffer));
+	if (!ent) {
+		last_errno = errno;
+		ret = ECGOTHER;
+		goto err;
+	}
+
+	mntopt = hasmntopt(ent, mnt_option);
```
drakenclimber, line 1362: This [commit](a18ae861e77f106517df203e30a810ba59b082d9) added smarts to ensure that `hasmntopt()` matched the whole opt string and not just a substring.  Do we want to do that here?

drakenclimber (changes requested):

drakenclimber:
@0n-s, do you want credit for reporting this bug (beyond listing your github username)?  If so, please provide your name and email and we'll add you to the commit like this:

`Reported-by: Your Name <awesome-contributor@libcgroup.notrealemail>`

drakenclimber:
I think these changes are on the right track, @kamalesh-babulal.  Do we have a test (and test machine) that can verify we've gotten this right?

src/api.c
drakenclimber: A clearer way to state what I wrote above.  I think there should be three clear returns from this function:
* `mnt_option` WAS found
* `mnt_option` WASN'T found
* error

src/api.c
```diff
@@ -1321,6 +1321,60 @@ static void cgroup_free_cg_mount_table(void)
...
 
+/*
+ * Parses the mount options of the given mount point and checks for the
+ * option in the list of mount options.
+ * @mnt: Mount point name to search for mount points.
+ * @mnt_option: Mount option to be searched.
+ *
+ * Returns 0, in case of success and ECGOTHER on failure.
+ */
+static int check_mount_point_opt(char *mnt, char *mnt_option)
```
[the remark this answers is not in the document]
kamalesh-babulal: @drakenclimber Thanks for reviewing the patch.  Agreed, will rename it.

src/api.c
kamalesh-babulal: My bad, impatient me did not read the `setmntent()` man page thoroughly. Sure will change it.

src/api.c
kamalesh-babulal: Great suggestion, yeah we might need something like that.

kamalesh-babulal:
**V2 Changes**
Addressed the following review comments from @drakenclimber 
- Use `endmntent()` instead of `fclose()` on `FILE*` returned by `setmntent()`.
- Add additional checks to the `hasmntopt()` logic, as it matches the wrong option, if `mnt_opt` passed is a subset of one of mount points.
- In `check_mount_point_opt()` add `is_set` function argument.

0n-s:
> @0n-s, do you want credit for reporting this bug (beyond listing your github username)?

Just my username should be fine, thanks. :-)

kamalesh-babulal:
> I think these changes are on the right track, @kamalesh-babulal. Do we have a test (and test machine) that can verify we've gotten this right?

@drakenclimber I need to add a test case. I have been testing it over my personal VM, with the following scenarios:
1. Cgroup Legacy Mode + `/proc` mounted without `subset=pid`
2. Cgroup Legacy Mode + `/proc` mounted with `subset=pid`
3. Cgroup Hybrid Mode + `/proc` mounted without `subset=pid`
4. Cgroup Hybrid Mode + `/proc` mounted with `subset=pid`
5. Cgroup Unified Mode + `/proc` mounted without `subset=pid`
6. Cgroup Unified Mode + `/proc` mounted with `subset=pid`

drakenclimber (approved):

drakenclimber:
Pushed to main in commits 16be308b037b and c83cb5a2d2ff.  Pushed to release-2.0 in commits 1686eb6f24b3 and d7cd75e24c27.  Pushed to release-3.0 in commits 860ac4456c6a and 050dd1c0abb1.

Thanks, @0n-s and @kamalesh-babulal

PR #239, opened 2022-08-24
github/workflows: add support for testing different cgroup setups

kamalesh-babulal:
Add support to run the Github workflow on three parallel runners, each
of them have a unique cgroup setup:
cgroup v1 only (legacy), cgroup v1/v2 (hybrid), cgroup v2 only (unified).

With this change, we should be able to run every patch against all
combination of cgroup setups.

Suggested-by: Tom Hromatka <tom.hromatka@oracle.com>

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -25,9 +25,27 @@ on: ["push", "pull_request"]
 jobs:
   # Thanks to github user @martyrs
   # https://github.community/t/how-to-properly-clean-up-self-hosted-runners/128909/3
-  cleaner:
-    name: Delete Self-Hosted Runner Workspaces
-    runs-on: self-hosted
+  cleaner_legacy:
+    name: Delete Self-Hosted Runner Workspaces - legacy
+    runs-on: cgroup-legacy
+    steps:
+      - name: Delete workspace path
+        run: |
+          echo "Cleaning up previous run"
+          rm -rf "${{ github.workspace }}"
+
+  cleaner_hybrid:
+    name: Delete Self-Hosted Runner Workspaces - hybrid
+    runs-on: cgroup-hybrid
+    steps:
+      - name: Delete workspace path
+        run: |
+          echo "Cleaning up previous run"
+          rm -rf "${{ github.workspace }}"
+
+  cleaner_unified:
+    name: Delete Self-Hosted Runner Workspaces - unified
+    runs-on: cgroup-unified
     steps:
       - name: Delete workspace path
         run: |
@@ -235,10 +253,10 @@ jobs:
         name: Cgroup v1v2 Code Coverage
         path: lcov.*
 
-  functionaltestsv2:
-    name: Cgroup v2 Functional Tests
-    needs: [cleaner]
-    runs-on: self-hosted
+  functionaltestsv2_legacy:
+    name: Cgroup v2 Functional Tests - legacy
+    needs: [cleaner_legacy]
+    runs-on: cgroup-legacy
 
     steps:
     - uses: actions/checkout@v2
@@ -263,7 +281,7 @@ jobs:
       if: ${{ always() }}
       uses: actions/upload-artifact@v2
       with:
-        name: Cgroup v2 test logs
+        name: Cgroup v2 test logs - legacy
         path: tests/ftests/*.log
     - name: Collate code coverage results
       uses: ./.github/actions/code-coverage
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
Looks like there are troubles with the connectivity (probably the infrastructure) with one of the VM.  We need to wait until it gets resolved.

.github/workflows/continuous-integration.yml
```diff
@@ -235,10 +253,102 @@ jobs:
...
+        name: Cgroup v2 test logs
+        path: tests/ftests/*.log
+    - name: Collate code coverage results
+      uses: ./.github/actions/code-coverage
+    - name: Upload code coverage results
+      uses: coverallsapp/github-action@master
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        path-to-lcov: ./lcov.total
+        flag-name: "Cgroup v2 Functional Tests"
```
drakenclimber, line 293: You'll probably want to add legacy to the `flag-name`

.github/workflows/continuous-integration.yml
```diff
@@ -235,10 +253,102 @@ jobs:
...
+    - name: Display test logs
+      if: ${{ always() }}
+      run: |
+        cat tests/ftests/ftests.sh.log
+        cat tests/ftests/ftests-nocontainer.sh.log
+    - name: Archive test logs
+      if: ${{ always() }}
+      uses: actions/upload-artifact@v2
+      with:
+        name: Cgroup v2 test logs
```
drakenclimber, line 284: You'll probably want to add legacy to the `name`

.github/workflows/continuous-integration.yml
```diff
@@ -235,10 +253,102 @@ jobs:
...
```
drakenclimber, line 330: You'll probably want to add hybrid to the `name`

.github/workflows/continuous-integration.yml
```diff
@@ -235,10 +253,102 @@ jobs:
...
+        path: tests/ftests/*.log
+    - name: Collate code coverage results
+      uses: ./.github/actions/code-coverage
+    - name: Upload code coverage results
+      uses: coverallsapp/github-action@master
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        path-to-lcov: ./lcov.total
+        flag-name: "Cgroup v2 Functional Tests"
```
drakenclimber, line 339: You'll probably want to add hybrid to the `flag-name`

drakenclimber: Github won't let me comment on unchanged lines, but you'll also want to add `unified` to the unified `name` and `flag-name` variables below

.github/workflows/continuous-integration.yml
```diff
@@ -25,9 +25,27 @@ on: ["push", "pull_request"]
 jobs:
   # Thanks to github user @martyrs
   # https://github.community/t/how-to-properly-clean-up-self-hosted-runners/128909/3
-  cleaner:
-    name: Delete Self-Hosted Runner Workspaces
-    runs-on: self-hosted
```
drakenclimber, line 30: Stupid question.  If we leave this as `self-hosted` won't it run on all machines that are tagged as self-hosted?  Then we don't need to duplicate this code three times

drakenclimber: Oh, I see... what you did forces it to run on all three machines.

I think an easier way may be to convert this to an action and then invoke it as follows from `functionaltestsv2_cgroup_legacy`, `functionaltestsv2_cgroup_hybrid`, etc.:

```
- name: Clean the workspace
  uses: ./.github/actions/clean-workspace
```

drakenclimber: One more random thought.  It looks like Github added the ability to run a script before or after a job.  They specifically cite a cleanup example
https://docs.github.com/en/actions/hosting-your-own-runners/running-scripts-before-or-after-a-job

.github/workflows/continuous-integration.yml
kamalesh-babulal: @drakenclimber Thank you so much for the review comments. I will address all of the comments in the next revision.

.github/workflows/continuous-integration.yml
kamalesh-babulal: @drakenclimber Thanks for the great suggestions. I tried both of your suggestions but was not successful with either of them :(.  The `run a script` approach, is not documented well enough for me :( and I tried different combinations and ideas to export/trigger `ACTIONS_RUNNER_HOOK_JOB_STARTED` but was not able to get it working. 
The other approach of using `./.github/actions/clean-workspace` has one issue, the git is cloned once and every action is checked out from the cloned tree, so running the action to delete the workspace will delete the tree, and hence the tree is not available for other steps, ending up with the failed job. I also tried using the cleanup action as the last step, thinking it will safe to do so, as we do not need the tree any more and it too failed.
I have another idea, of integrating the `cleaner` step, as the first one to run, something like below:
```

**Comment by kamalesh-babulal** [comment]
**V2 Changes**
- Addressed the review comments from @drakenclimber to add the right runner labels to the job steps.

File: .github/workflows/continuous-integration.yml (hunk 5 above)
Comment (drakenclimber) [reply .github/workflows/continuous-integration.yml:30]: I got it working on our runners.  I couldn't get the `.env` file method to work, but it did work by setting the environment variable `$ACTIONS_RUNNER_HOOK_JOB_STARTED` to point to a script to clean the workspace.

Here's the script I'm using
```
#!/bin/bash

echo "Deleting $GITHUB_WORKSPACE" | sudo tee /dev/kmsg
rm -rf $GITHUB_WORKSPACE
mkdir -p $GITHUB_WORKSPACE
```

drakenclimber:
Pushed to main in commit 995202c4f53fe6c7d5b90e95c58f77c7a4280d71.  Pushed to release-3.0 in commit 66d7c3154889c58d879284ff716cf6b7d7f5fb39.  _Not_ pushed to release-2.0

PR #251, opened 2022-09-17
tools/cgget: fix segfault in get_cv_value() [v2.0.2]

kamalesh-babulal:
cgget segfaults on v2.0.2 branch, with:
```
cgget: cannot find controller 'incal' in group '016cgget'
Fatal error: glibc detected an invalid stdio handle
Aborted (core dumped)
```
It was caught by `ftests/016-cgget-invalid_options.py` on Ubuntu 22.04 (gcc version 11.2.0),
a simple reproducer on the v2.0.2 branch:
`$ sudo ./src/tools/cgget -n -v -r invalid.setting 016cgget` assuming 016cgget cgroup exists.

It is due to the invalid controller name passed to the `cgroup_read_value_begin()`,
which returns failure and callee `get_cv_value()` in the error clean up path, does a
`fclose(handle)`. `If (handle != NULL)` succeeds because it's uninitialized and has
some garbage value. Fix this by initializing the handle to NULL.

### DIFF
--- a/src/tools/cgget.c
+++ b/src/tools/cgget.c
@@ -462,7 +462,7 @@ static int get_cv_value(struct control_value * const cv,
 {
 	bool is_multiline = false;
 	char tmp_line[LL_MAX];
-	void *handle, *tmp;
+	void *tmp, *handle = NULL;
 	int ret;
 
 	ret = cgroup_read_value_begin(controller_name, cg_name, cv->name,


drakenclimber:
Pushed to main in commit c8af263a8802b48a3fdc0de68e6408d72619e172.  Pushed to release-2.0 in commit 91f1fb6e5cfa3e23d533775edbf7243cf0262a71.  Pushed to release-3.0 in commit 10f973e09581ca5565c9a54d796094dfcca8fdac

Thanks, @kamalesh-babulal

PR #254, opened 2022-09-23
tools/cgdelete: add check for the cmdline arguments [v2.0.2]

kamalesh-babulal:
Like other tools, throw an error and exit, if the minimum required arguments are not passed.


drakenclimber (approved):

drakenclimber:
Pushed to release-2.0 in commit 0fc4d286cdc12959d2c7ed3f451ce2756e7fd179.  Thanks, @kamalesh-babulal

PR #255, opened 2022-09-23
tools/cgdelete: add check for the cmdline arguments [v3.0+]

kamalesh-babulal:
Like other tools, throw an error and exit if the minimum required arguments are not passed.


kamalesh-babulal:
@drakenclimber If you like the idea proposed at #256, please do not merge this commit. I will redo the patch with `exit(129);`

drakenclimber (approved):

PR #256, opened 2022-09-25
[RFC] tools/: use exit code 129 on invalid usage

kamalesh-babulal:
The exit code 1 refers to general error and using -1 wraps to
255, both codes do not signify bad arguments. Let's use exit
code 129 for hinting to the users about invalid usage.  This
patch series changes the exit code to 129, when users pass
invalid/bad parameter to all of the tools and also introduces
the changes to `ftests/016`, `ftests/027` testcases, those test
the tools by passing invalid parameters and expect exit codes
return when passed with invalid parameters.


kamalesh-babulal:
** v2 Change **
- Fix the commit subject on patch `0005` (changed from `tools/cgget` to `tools/cgclassify`)

drakenclimber:
This seems like a good idea to me.  One small nitpick - should we add a constant somewhere and use that rather than hard-coding `129` everywhere?

drakenclimber:
FYI - I pushed the cgdelete arg count fix to release-2.0.  release-2.0 won't get this `129` change, so we should put in the simpler fix there.

kamalesh-babulal:
@drakenclimber Thanks for the review. Does some thing like `EBADARGS` or `EINVALARGS` as the constant?

kamalesh-babulal:
> FYI - I pushed the cgdelete arg count fix to release-2.0. release-2.0 won't get this 129 change, so we should put in the simpler fix there.

Thanks for merging it into release-2.0. Sure, will add it to this patchseries.

drakenclimber:
> @drakenclimber Thanks for the review. Does some thing like `EBADARGS` or `EINVALARGS` as the constant?

Either sounds good to me.

kamalesh-babulal:
**V3 Changes**
- Introduce constant `EXIT_BADARGS`, which will be used by the tools to exit on bad parameters instead of exit code 129.
- Folded patch #255 into this patch series.

src/tools/cgsnapshot.c
```diff
@@ -767,20 +767,24 @@ int main(int argc, char *argv[])
 		output_f = stdout;
 
 	/* blacklkist */
-	if (flags & FL_BLACK) {
+	if (flags & FL_BLACK)
```
drakenclimber, line 770: nitpick - I thought the kernel syntax recommendation in a case like this was to keep the braces.  Am I old and out of touch? :)

https://www.kernel.org/doc/html/v4.10/process/coding-style.html

src/tools/cgsnapshot.c
```diff
@@ -767,20 +767,24 @@ int main(int argc, char *argv[])
...
 		ret  = load_list(bl_file, &black_list);
-	} else {
+	else {
 		/* load the blacklist from the default location */
 		ret  = load_list(BLACKLIST_CONF, &black_list);
 	}
-	if (ret != 0)
+	if (ret != 0) {
+		ret = ret ? EXIT_BADARGS : 0;
```
drakenclimber, line 777: Won't this always evaluate to `EXIT_BAD_ARGS` since in the `if` line above we checked if `ret != 0`

src/tools/cgsnapshot.c
```diff
@@ -767,20 +767,24 @@ int main(int argc, char *argv[])
...
 		goto finish;
+	}
 
 	/* whitelist */
 	if (flags & FL_WHITE)
 		ret = load_list(wl_file, &white_list);
-	if (ret != 0)
+	if (ret != 0) {
+		ret = ret ? EXIT_BADARGS : 0;
```
drakenclimber, line 785: Same comment as above

src/tools/cgset.c
```diff
@@ -88,7 +88,7 @@ STATIC int parse_r_flag(const char * const program_name, const char * const name
 	buf = strtok(copy, "=");
 	if (buf == NULL) {
 		err("%s: wrong parameter of option -r: %s\n", program_name, optarg);
-		ret = -1;
+		ret = EXIT_BADARGS;
```
drakenclimber, line 91: Good find.  I grepped the tools folder for calls to `usage()`, and that missed many invalid parameter cases like this.

drakenclimber:
I had a couple silly comments, @kamalesh-babulal, but I think the changes look good.

src/tools/cgsnapshot.c
kamalesh-babulal: @drakenclimber Thanks for the review, it's a long patch set. I am wondering why the `checkpatch.pl` did not flag it. I will try to figure out what went wrong. Agreed, will fix it.

src/tools/cgsnapshot.c
kamalesh-babulal: Good catch :). Sure will change it here and other places too.

kamalesh-babulal:
**V4 Changes**
- Fixed the review comments from @drakenclimber in the `cgsnapshot.c` file.

drakenclimber (approved):

drakenclimber:
Pushed to main and release-3.0.  As mentioned previously, let's not push this to release-2.0 as it's too large of a change for that branch.

drakenclimber:
Thanks, @kamalesh-babulal

PR #260, opened 2022-10-03
[RFC] Introduce a new API to find the cgroup setup mode

kamalesh-babulal:
This patchset introduces a new API `cgroup_setup_mode()` to
detect the current cgroup setup mode (Legacy/Unified/Hybrid).
The setup will depend on the Linux Kernel's grub command line
the system/VM is booted with.

One use case for this is, the users can enable or disable features
in their applications depending upon the cgroup setup. Like some
controllers are only available on cgroup v2, so they might need to
set/get settings for those available controllers only.

The second patch, adds a new switch (-m) to `cgget`, that will print
the current cgroup mode the system/VM has booted it.

### DIFF
--- a/include/libcgroup/groups.h
+++ b/include/libcgroup/groups.h
@@ -23,6 +23,13 @@ enum cg_version_t {
 	CGROUP_DISK = 0xFF,
 };
 
+enum cg_setup_mode_t {
+	CGROUP_MODE_UNKNOWN = 0,
+	CGROUP_MODE_LEGACY,
+	CGROUP_MODE_HYBRID,
+	CGROUP_MODE_UNIFIED,
+};
+
 /**
  * Flags for cgroup_delete_cgroup_ext().
  */
@@ -635,6 +642,12 @@ int cgroup_list_mount_points(const enum cg_version_t cgrp_version,
  */
 int cgroup_get_controller_version(const char * const controller,
 				  enum cg_version_t * const version);
+
+/**
+ * Finds the current group setup mode (legacy/unified/hybrid).
+ * Returns unknown of failure and setup mode on success.
+ */
+enum cg_setup_mode_t cgroup_setup_mode(void);
 
 /**
  * @}
--- a/src/api.c
+++ b/src/api.c
@@ -43,6 +43,7 @@
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/vfs.h>
 
 #include <linux/un.h>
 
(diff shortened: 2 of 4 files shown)


kamalesh-babulal:
Will contribute the test case based on the accepted API interface.

src/api.c
```diff
@@ -6118,3 +6119,65 @@ const struct cgroup_library_version *cgroup_version(void)
...
+ * Finds the current cgroup setup mode (legacy/unified/hybrid).
+ * Returns unknown of failure and setup mode on success.
+ */
+enum cg_setup_mode_t cgroup_setup_mode(void)
+{
+/* FS codes refered from include/magic.h */
+#define CGROUP2_SUPER_MAGIC     0x63677270
+#define TMPFS_MAGIC             0x01021994
+
+	const char *cgrp_sysfs_unified = "/sys/fs/cgroup/unified/";
```
drakenclimber, line 6133: We have some users that don't use systemd.  Should we check for the existence of `/sys/fs/cgroup/`.  If it exists, use this shortcut.  Otherwise, look through the cg_mount_table.

src/libcgroup.map
```diff
@@ -146,4 +146,5 @@ CGROUP_3.0 {
 	cgroup_cgxset;
 	cgroup_version;
 	cgroup_list_mount_points;
+	cgroup_setup_mode;
```
drakenclimber, line 149: Nitpick - we should probably add a comment above this method.  (The map file doesn't support naming like `CGROUP_3.0.1`.  Not sure why.)

```
cgroup_list_mount_points;

/* libcgroup 3.0.1 */
cgroup_setup_mode;
```

drakenclimber (commented):
I had a couple minor-ish comments, but I think this will be helpful.  Looks like a good start, @kamalesh-babulal.  Thanks.

src/api.c
drakenclimber: Or is this a "systemd-only" interface?  If `/sys/fs/cgroup` doesn't exist, should we just return an error?

src/libcgroup.map
kamalesh-babulal: @drakenclimber Thanks for the review. Sure, will include it in the next version of the patch. It is interesting why the naming is supported.

src/api.c
kamalesh-babulal: >  We have some users that don't use systemd. Should we check for the existence of /sys/fs/cgroup/. If it exists, use this shortcut. Otherwise, look through the cg_mount_table. Or is this a "systemd-only" interface? If /sys/fs/cgroup doesn't exist, should we just return an error?

I was thinking of the API from the view from systemd only. Agreed, I will re-think and come up with a better approach based on your suggestion of using the `cg_mount_table`.

kamalesh-babulal: > Just beware the [directory in sysfs exists](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/cgroup/cgroup.c?id=833477fce7a14d43ae4c07f8ddc32fa5119471a2#n6000) regardless of systemd. What is different whether/what is mounted at that path. Mountinfo looks a more robust way in general landscape.

@Werkov Thank you for the review and testing of the patchset.  Yeah, the kernel creates the `/sys/fs/cgroup` during `cgroup_init()`. I will try to re-implement it using `cg_mount_table` that already captures the details about the mount points and also make it more robust by detecting all allowed cgroup setup combinations, those get influenced by `cgroup` kernel parameters.

kamalesh-babulal:
@drakenclimber @Werkov The latest version relies on the `cg_mount_table()` to detect the mount setup. Currently, the `cg_mount_table()` is populated with `V1/V2 mount points`, where the v1 mount points, have controllers attached or might be named hierarchy, and in the case of V2 mount point have controllers attached. The empty V2 mount points are not part of the table and they are tracked in `cg_cgroup_v2_empty_mount_paths`.  The algorithm to detect the cgroup is:

1. loop through the `cg_mount_table` and if its cgroup v1 the resepctive bit mask is set and same with cgroup v2

include/libcgroup/groups.h
```diff
@@ -23,6 +23,13 @@ enum cg_version_t {
 	CGROUP_DISK = 0xFF,
 };
 
+enum cg_setup_mode_t {
+	CGROUP_UNKNOWN = 0,
```
drakenclimber, line 27: To avoid confusion and collisions with `enum cg_version_t` would it be better to name these something like:
```
CGROUP_MODE_UNK,
CGROUP_MODE_LEGACY,
CGROUP_MODE_HYBRID,
CGROUP_MODE_UNFIED,
```

src/api.c
```diff
@@ -6118,3 +6119,49 @@ const struct cgroup_library_version *cgroup_version(void)
...
+enum cg_setup_mode_t cgroup_setup_mode(void)
+{
+#define CGROUP2_SUPER_MAGIC	0x63677270
+#define CGROUP_SUPER_MAGIC	0x27E0EB
+
+	unsigned int cg_setup_mode_bitmask = 0U;
+	enum cg_setup_mode_t setup_mode;
+	struct statfs cgrp_buf;
+	int i, ret = 0;
+
```
drakenclimber, line 6136: Now that we are walking `cg_mount_table`, do we need a block like this?
```
        if (!cgroup_initialized) {
                return ECGROUPNOTINITIALIZED;
        }    
```

src/api.c
```diff
@@ -6118,3 +6119,49 @@ const struct cgroup_library_version *cgroup_version(void)
...
+	setup_mode = CGROUP_UNKNOWN;
+
+	pthread_rwlock_wrlock(&cg_mount_table_lock);
+	for (i = 0; cg_mount_table[i].name[0] != '\0'; i++) {
+		ret = statfs(cg_mount_table[i].mount.path, &cgrp_buf);
+		if (ret) {
+			setup_mode = CGROUP_UNKNOWN;
+			cgroup_dbg("Failed to get stats of '%s'\n", cg_mount_table[i].mount.path);
```
drakenclimber, line 6144: nitpick - should we promote this to `cgroup_err()` since this is a failure case?

drakenclimber (commented):
I had a few silly comments, but I think this looks good, @kamalesh-babulal.  Thanks for the patchset

include/libcgroup/groups.h
kamalesh-babulal: @drakenclimber Apology for the late reply was caught up with other projects. The suggested renames look better will change them.

src/api.c
kamalesh-babulal: Sure, lets error out in this unlikely event().

kamalesh-babulal:
**V3 Changes**
- Addressed review comments from @drakenclimber

drakenclimber:
[v2](https://github.com/libcgroup/libcgroup/pull/263) of this patchset was pushed to main and release-3.0.  Not pushed to release-2.0.

Thanks, @kamalesh-babulal.

PR #263, opened 2022-10-19
Introduce a new API to find the cgroup setup mode (v2)

drakenclimber:
This patchset introduces a new API `cgroup_setup_mode()` to
detect the current cgroup setup mode (Legacy/Unified/Hybrid).
The setup will depend on the Linux Kernel's grub command line
the system/VM is booted with.

One use case for this is, the users can enable or disable features
in their applications depending upon the cgroup setup. Like some
controllers are only available on cgroup v2, so they might need to
set/get settings for those available controllers only.

The second patch, adds a new switch (-m) to `cgget`, that will print
the current cgroup mode the system/VM has booted it.

Changes from v1 -> v2:
* Add python bindings
* Add functional test
* Address review comments from v1

v1 patchset is available here:
https://github.com/libcgroup/libcgroup/pull/260

### DIFF
--- a/include/libcgroup/groups.h
+++ b/include/libcgroup/groups.h
@@ -23,6 +23,13 @@ enum cg_version_t {
 	CGROUP_DISK = 0xFF,
 };
 
+enum cg_setup_mode_t {
+	CGROUP_MODE_UNK = 0,
+	CGROUP_MODE_LEGACY,
+	CGROUP_MODE_HYBRID,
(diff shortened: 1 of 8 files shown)


kamalesh-babulal (approved):
@drakenclimber Thank you so much, for working on the Python bindings/Testcases.

drakenclimber:
Pushed to main and release-3.0.  Not pushed to release-2.0.

Thanks, @kamalesh-babulal.

PR #266, opened 2022-10-27
Add Python lint to the Github Workflow [v3.0+]

kamalesh-babulal:
The Python lint in the GitHub workflow, was previously part of the `libcgroup-test`
repo, which hosted most of the python source code.  Now that `libcgroup-test`
is part of `libcgroup` repo, let's add the lint workflow here too. Also, the `flake8`
Python lint tool, allows us to tweak the checks by customizing and extending the
rules the lint should operate. To begin with, let's add our own rule to allow a
100-character width in the sources.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 100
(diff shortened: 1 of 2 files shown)


drakenclimber (approved):

drakenclimber:
Thanks, @kamalesh-babulal.  Pushed to main in commits e211970c9ee2 and 5eee3084f292.  Pushed to release-3.0 in commits 2622cea9ad53 and 1a70a33a8131.  Not pushed to release-2.0.

PR #269, opened 2022-11-07
[RFC] Add systemd configuration support to cgconfig.conf

kamalesh-babulal:
This patch series introduces systemd configurations to `cgconfig.conf`
and support to parse them. This will help the users to create `systemd slice/scope`
when system boots, and also allow them to set the resource allocations to the created
`slice/scope`.  A sample configuration would look like this:
```
systemd {
        slice = testing;
        scope = testing;
}

systemd {
        slice = testing;
        scope = testing1;
        setdefault = yes;
}

group testing.slice {
        cpu {
               cpu.weight = 10;
        }
}
```
and it also adds support to the tools to consider the default slice as the cgroup root
for all operations performed by the tool, instead of the controller mount point.  This
will help users on working on the default `slice/scope` hierarchy they created as the
new cgroup root .  The user can also ignore the default hierarchy setting by using the
new switch `-b` added to all of the tools.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -97,6 +97,11 @@ struct cg_mount_table_s cg_mount_table[CG_CONTROLLER_MAX];
 /* Cgroup v2 mount paths, with empty controllers */
 struct cg_mount_point *cg_cgroup_v2_empty_mount_paths;
 
+#ifdef WITH_SYSTEMD
+/* Default systemd path name. Length: <name>.slice/<name>.scope */
+char systemd_default_cgroup[FILENAME_MAX * 2 + 1];
+#endif
+
 const char * const cgroup_strerror_codes[] = {
 	"Cgroup is not compiled in",
 	"Cgroup is not mounted",
--- a/src/api.c
+++ b/src/api.c
@@ -1680,7 +1680,24 @@ static char *cg_concat_path(const char *pref, const char *suf, char *path)
 /* path value have to have size at least FILENAME_MAX */
 char *cg_build_path_locked(const char *name, char *path, const char *type)
 {
-	int i, ret;
+	/*
+	 * len is the allocation size for path, that stores:
+	 * cg_mount_table[i].mount.path + '/' + cg_namespace_table[i] + '/'
+	 */
+	int i, ret, len = (FILENAME_MAX * 2) + 2;
+	char *_path;
+
+	/*
+	 * Recent gcc are unhappy when sizeof(dest) <= sizeof(src) with
+	 * snprintf()'s.  Alternative is to use multiple strncpy()/strcat(),
+	 * work around it by allocating large temporary buffer _path and
+	 * copying the constructed _path into path.
+	 */
+	_path = malloc(len);
+	if (!_path) {
+		cgroup_err("Failed to allocate memory for _path\n");
+		return NULL;
+	}
 
 	/*
 	 * If no type is specified, and there's a valid cgroup v2 mount, then
@@ -1689,21 +1706,26 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
 	 * any controller.
 	 */
 	if (!type && strlen(cg_cgroup_v2_mount_path) > 0) {
-		ret = snprintf(path, FILENAME_MAX, "%s/", cg_cgroup_v2_mount_path);
+		ret = snprintf(_path, len, "%s/", cg_cgroup_v2_mount_path);
 		if (ret >= FILENAME_MAX)
-			cgroup_dbg("filename too long: %s/", cg_cgroup_v2_mount_path);
+			cgroup_dbg("filename too long: %s", _path);
+
+		strncpy(path, _path, FILENAME_MAX - 1);
+		path[FILENAME_MAX - 1] = '\0';
 
 		if (name) {
 			char *tmp;
 
 			tmp = strdup(path);
-			if (tmp == NULL)
-				return NULL;
+			if (tmp == NULL) {
+				path = NULL;
+				goto out;
+			}
 
 			cg_concat_path(tmp, name, path);
 			free(tmp);
 		}
-		return path;
+		goto out;
 	}
 
 	for (i = 0; cg_mount_table[i].name[0] != '\0'; i++) {
@@ -1717,23 +1739,17 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
 		    (type && strcmp(type, CGROUP_FILE_PREFIX) == 0 &&
 		     cg_mount_table[i].version == CGROUP_V2)) {
 
-			if (cg_namespace_table[i]) {
-				ret = snprintf(path, FILENAME_MAX, "%s/%s/",
-						cg_mount_table[i].mount.path,
-						cg_namespace_table[i]);
-				if (ret >= FILENAME_MAX) {
-					cgroup_dbg("filename too long:%s/%s/",
-						   cg_mount_table[i].mount.path,
-						   cg_namespace_table[i]);
-				}
-			} else {
-				ret = snprintf(path, FILENAME_MAX, "%s/",
-					       cg_mount_table[i].mount.path);
-				if (ret >= FILENAME_MAX) {
-					cgroup_dbg("filename too long:%s/",
-						   cg_mount_table[i].mount.path);
-				}
-			}
+			if (cg_namespace_table[i])
+				ret = snprintf(_path, len, "%s/%s/", cg_mount_table[i].mount.path,
+					       cg_namespace_table[i]);
+			else
+				ret = snprintf(_path, len, "%s/", cg_mount_table[i].mount.path);
+
+			if (ret >= FILENAME_MAX)
+				cgroup_dbg("filename too long: %s", _path);
+
+			strncpy(path, _path, FILENAME_MAX - 1);
+			path[FILENAME_MAX - 1] = '\0';
 
 			if (name) {
 				char *tmp;
@@ -1745,10 +1761,16 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
 				cg_concat_path(tmp, name, path);
 				free(tmp);
 			}
-			return path;
+			goto out;
 		}
 	}
-	return NULL;
+	path = NULL;
+
+out:
+	if (_path)
+		free(_path);
+
+	return path;
 }
 
 char *cg_build_path(const char *name, char *path, const char *type)
--- a/src/api.c
+++ b/src/api.c
@@ -1690,8 +1690,34 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
 	 * cg_mount_table[i].mount.path + '/' + cg_namespace_table[i] + '/'
 	 */
 	int i, ret, len = (FILENAME_MAX * 2) + 2;
-	char *_path;
+	char *tmp_systemd_default_cgroup, *_path = NULL;
 
+	/*
+	 * systemd_default_cgroup can't be clobbered.   The user may pass
+	 * multiple cgroups, hence use temporary variable for manipulations
+	 * for example:
+	 * cgget -g cpu:/ -g cpu:cgrp1 -g cpu:/cgrp2
+	 */
+	tmp_systemd_default_cgroup = calloc(1, (sizeof(char) * len));
+	if (!tmp_systemd_default_cgroup) {
+		cgroup_err("Failed to allocate memory for tmp_systemd_default_cgroup\n");
+		goto out;
+	}
+
+#ifdef WITH_SYSTEMD
+	/*
+	 * If the user specifies the name as /<cgroup-name>, they are
+	 * effectively overriding the systemd_default_cgroup but if the name
+	 * is "/", the cgroup root path is systemd_default_cgroup
+	 */
+	if (strlen(systemd_default_cgroup) && name && name[0] == '/' && name[1] != '\0')
+		tmp_systemd_default_cgroup[0] = '\0';
+	else
+		snprintf(tmp_systemd_default_cgroup, len, "%s/", systemd_default_cgroup);
+
+	/* allocate more space for systemd_default_cgroup + '/' */
+	len += (FILENAME_MAX + 1);
+#endif
 	/*
 	 * Recent gcc are unhappy when sizeof(dest) <= sizeof(src) with
 	 * snprintf()'s.  Alternative is to use multiple strncpy()/strcat(),
@@ -1701,7 +1727,7 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
 	_path = malloc(len);
 	if (!_path) {
 		cgroup_err("Failed to allocate memory for _path\n");
-		return NULL;
+		goto out;
 	}
 
 	/*
@@ -1711,7 +1737,8 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
 	 * any controller.
 	 */
 	if (!type && strlen(cg_cgroup_v2_mount_path) > 0) {
-		ret = snprintf(_path, len, "%s/", cg_cgroup_v2_mount_path);
+		ret = snprintf(_path, len, "%s/%s", cg_cgroup_v2_mount_path,
+			       tmp_systemd_default_cgroup);
 		if (ret >= FILENAME_MAX)
 			cgroup_dbg("filename too long: %s", _path);
 
@@ -1745,10 +1772,11 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
 		     cg_mount_table[i].version == CGROUP_V2)) {
 
 			if (cg_namespace_table[i])
-				ret = snprintf(_path, len, "%s/%s/", cg_mount_table[i].mount.path,
-					       cg_namespace_table[i]);
+				ret = snprintf(_path, len, "%s/%s%s/", cg_mount_table[i].mount.path,
+					       tmp_systemd_default_cgroup, cg_namespace_table[i]);
 			else
-				ret = snprintf(_path, len, "%s/", cg_mount_table[i].mount.path);
+				ret = snprintf(_path, len, "%s/%s", cg_mount_table[i].mount.path,
+					       tmp_systemd_default_cgroup);
 
 			if (ret >= FILENAME_MAX)
 				cgroup_dbg("filename too long: %s", _path);
@@ -1774,6 +1802,9 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
 out:
 	if (_path)
 		free(_path);
+
+	if (tmp_systemd_default_cgroup)
+		free(tmp_systemd_default_cgroup);
 
 	return path;
 }
--- a/src/config.c
+++ b/src/config.c
@@ -2077,6 +2077,87 @@ static int config_create_slice_scope(char * const tmp_systemd_default_cgroup)
 err:
 	return 0;
 }
+
+/*
+ * This function builds the canonical path for the systemd default cgroup
+ * based on the cgroup setup mode and if the cgroup doesn't exist removes
+ * the systemd default cgroup file. This function expects the caller to be
+ * holding the systemd_default_cgroup_lock.
+ */
+static bool systemd_default_cgroup_exists(void)
+{
+	char path[FILENAME_MAX] = "\0";
+	char cgrp_procs_path[FILENAME_MAX + 13]; /* 13 = strlen(/cgroup.procs) */
+	bool exists = false;
+
+	switch (cgroup_setup_mode()) {
+	case CGROUP_MODE_HYBRID:
+		/*
+		 * check for empty cgroup v2, the most common usage in
+		 * the hybrid case.
+		 */
+		if (cg_build_path("", path, NULL))
+			break;
+	case CGROUP_MODE_UNIFIED:
+		/* fallthrough */
+	case CGROUP_MODE_LEGACY:
+		cg_build_path("", path, "cpu");
+		/* fallthrough */
+	case CGROUP_MODE_UNK:
+		break;
+	}
+
+	if (!strlen(path)) {
+		cgroup_dbg("Unable to form canonical path for %s", systemd_default_cgroup);
+		goto err;
+	}
+
+	snprintf(cgrp_procs_path, sizeof(cgrp_procs_path), "%s/cgroup.procs", path);
+	if (access(cgrp_procs_path, F_OK)) {
+		/*
+		 * systemd has already removed the scope cgroup, hence delete
+		 * the file too.
+		 */
+		cgroup_dbg("%s doesn't exists, deleting %s", path, systemd_default_cgroup_file);
+		unlink(systemd_default_cgroup_file);
+		goto err;
+	}
+
+	exists = true;
+err:
+	return exists;
+}
+
+void cgroup_set_default_systemd_cgroup(void)
+{
+	FILE *systemd_def_cgrp_f;
+	size_t len;
+
+	pthread_rwlock_wrlock(&systemd_default_cgroup_lock);
+
+	systemd_def_cgrp_f = fopen(systemd_default_cgroup_file, "r");
+	if (!systemd_def_cgrp_f) {
+		cgroup_dbg("Unable to read %s ", systemd_default_cgroup_file);
+		goto err;
+	}
+
+	len = fread(systemd_default_cgroup, sizeof(char), FILENAME_MAX, systemd_def_cgrp_f);
+	fclose(systemd_def_cgrp_f);
+	/* at the minimum it should be <n>.slice/<n>.scope == 15 */
+	if (len < 15) {
+		cgroup_dbg("Malformed systemd default cgroup %s", systemd_default_cgroup);
+		goto err;
+	}
+
+	if (systemd_default_cgroup_exists()) {
+		pthread_rwlock_unlock(&systemd_default_cgroup_lock);
+		return;
+	}
+err:
+	pthread_rwlock_unlock(&systemd_default_cgroup_lock);
+	cgroup_dbg(", continuing without systemd default cgroup.\n", systemd_default_cgroup);
+	systemd_default_cgroup[0] = '\0';
+}
 #else
 int cgroup_add_systemd_opts(const char * const config, const char * const value)
 {
@@ -2089,4 +2170,6 @@ int cgroup_alloc_systemd_opts(const char * const config, const char * const valu
 }
 
 void cgroup_cleanup_systemd_opts(void) { }
+
+void cgroup_set_default_systemd_cgroup(void) { }
 #endif
--- a/src/config.c
+++ b/src/config.c
@@ -19,6 +19,7 @@
 
 #include <libcgroup.h>
 #include <libcgroup-internal.h>
+#include <libcgroup/systemd.h>
 
 #include <pthread.h>
 #include <assert.h>
@@ -90,6 +91,26 @@ static struct cgroup_string_list *template_files;
 
 /* Needed for the type while mounting cgroupfs. */
 #define CGROUP_FILESYSTEM "cgroup"
+
+#ifdef WITH_SYSTEMD
+/* Directory that holds dynamically created internal libcgroup files */
+static const char * const systemd_def_cgrp_file_dir = "/var/run/libcgroup/";
+
+/* File that stores the relative delegated systemd cgroup path */
+static const char * const systemd_default_cgroup_file = "/var/run/libcgroup/systemd";
+
+/* Lock that protects systemd_default_cgroup_file */
+static pthread_rwlock_t systemd_default_cgroup_lock = PTHREAD_RWLOCK_INITIALIZER;
+
+/*
+ * cgroup_systemd_opts is dynamically allocated, hence having head
+ * and tail helps in traversing.
+ */
+static struct cgroup_systemd_opts *cgroup_systemd_opts_head;
+static struct cgroup_systemd_opts *cgroup_systemd_opts_tail;
+
+static int config_create_slice_scope(char * const tmp_systemd_default_cgroup);
+#endif
 
 /*
  * NOTE: All these functions return 1 on success and not 0 as is the
@@ -971,6 +992,8 @@ static void cgroup_free_config(void)
 		config_template_table = NULL;
 	}
 	config_template_table_index = 0;
+
+	cgroup_cleanup_systemd_opts();
 }
 
 /**
@@ -1081,6 +1104,10 @@ static void cgroup_config_sort_groups(void)
  */
 int cgroup_config_load_config(const char *pathname)
 {
+#ifdef WITH_SYSTEMD
+	/* slice[FILENAME_MAX] + '/' + scope[FILENAME_MAX] */
+	char tmp_systemd_default_cgroup[FILENAME_MAX * 2 + 1] = "\0";
+#endif
 	int namespace_enabled = 0;
 	int mount_enabled = 0;
 	int error;
@@ -1126,11 +1153,34 @@ int cgroup_config_load_config(const char *pathname)
 	if (error)
 		goto err_mnt;
 
+#ifdef WITH_SYSTEMD
+	error = config_create_slice_scope(tmp_systemd_default_cgroup);
+	if (!error) {
+		error = ECGROUPPARSEFAIL;
+		goto err_mnt;
+	}
+#endif
+
 	cgroup_config_apply_default();
 	error = cgroup_config_create_groups();
 	cgroup_dbg("creating all cgroups now, error=%d\n", error);
 	if (error)
 		goto err_grp;
+
+#ifdef WITH_SYSTEMD
+	/*
+	 * Setting up systemd_default_cgroup, during creation
+	 * of slice/scopes, clobbers the path returned by cg_build_path(),
+	 * by appending the systemd_default_cgroup to it and
+	 * its required only after all of the slices/scopes are created.
+	 * The user might also set delegate in more than one systemd
+	 * delegate settings, in that case the last parsed one overwrites
+	 * the systemd_default_cgroup.
+	 */
+	if (strlen(tmp_systemd_default_cgroup))
+		snprintf(systemd_default_cgroup, sizeof(systemd_default_cgroup),
+			 "%s", tmp_systemd_default_cgroup);
+#endif
 
 	cgroup_free_config();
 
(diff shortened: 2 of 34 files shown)


Werkov:
The examples are slightly confusing (everything named `testing`, I'd suggest more distinctive names in the example for inner/leaf nodes), it's not clear what the `group testing.slice` refers to from the two sections above.

Also note, that [systemd delegation](https://systemd.io/CGROUP_DELEGATION/) is defined only for service and scopes, not slices (internal nodes). I understand your approach is similar (not identical) to *i-like-continents* scenario in that document.

kamalesh-babulal:
> The examples are slightly confusing (everything named testing, I'd suggest more distinctive names in the example for inner/leaf nodes), it's not clear what the group testing.slice refers to from the two sections above. Also note, that [systemd delegation](https://systemd.io/CGROUP_DELEGATION/) is defined only for service and scopes, not slices (internal nodes). I understand your approach is similar (not identical) to i-like-continents scenario in that document.

@Werkov Thank you for reviewing the patch.  If the example `cgconfig.conf` is used, the `systemd-cgls` output will look like:
```
[...]
│   └─ 728 /usr/bin/abrt-dump-journal-oops -fxtD
└─testing.slice 
  ├─testing.scope …
  │ └─ 986 libcgroup_systemd_idle_thread
  └─testing1.scope …
    └─ 987 libcgroup_systemd_idle_thread
```
I will try and come up with a better names for them and add more comments with references to the sample/example to make it more self-explanatory.

include/libcgroup/config.h
```diff
@@ -29,13 +29,57 @@ extern "C" {
...
  * See cgconfig.conf man page for format of the file.
  * @param pathname Name of the configuration file to load.
  */
 int cgroup_config_load_config(const char *pathname);
 
+/**
+ * Parse the systemd delegation settings from the configuration file
+ * and get the default delegation cgroup name, if set.
+ *
+ * @param pathname Name of the configuration file to load.
```
drakenclimber, line 53: How does this function behave if the default isn't set?

src/api.c
```diff
@@ -1752,6 +1774,33 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
 			if (name) {
 				char *tmp;
 
+				/*
+				 * If the user has passed absolute cgroup path,
+				 * avoid using systemd delegate slice as path
+				 * prefix.
+				 */
+				if (name[0] == '/') {
```
drakenclimber, line 1782: This feels kinda klunky.  So we build a path above, but if the first char is `/`, we'll override that path and put a new one in instead?

src/api.c
```diff
@@ -1752,6 +1774,33 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
...
+					if (cg_namespace_table[i]) {
```
drakenclimber, line 1783: Also, somewhat of a corner case, but tools like cgget used to allow the user to get settings in the root cgroup via `/` as the cgroup name.  How will that work here?

tests/Makefile.am
```diff
@@ -1,4 +1,5 @@
 DIST_SUBDIRS = ftests gunit
 if WITH_TESTS
-SUBDIRS = $(DIST_SUBDIRS)
+SUBDIRS = gunit
+#SUBDIRS = $(DIST_SUBDIRS)
```
drakenclimber, line 4: This feels like an oops :)

src/tools/cgget.c
```diff
@@ -746,6 +759,14 @@ int main(int argc, char *argv[])
 	if (ret)
 		goto err;
 
+	if (mode & MODE_SYSTEMD_DELEGATE) {
+		ret = cgroup_get_default_slice_name(CGCONFIG_CONF_FILE);
```
drakenclimber, line 763: What if the user is utilizing the cgconfigconfdir, `/etc/cgconfig.d`?

include/libcgroup/config.h
kamalesh-babulal: @drakenclimber Thank you so much for reviewing the patches. It will return the path without appending the default slice cgroup name.

src/config.c
```diff
@@ -608,6 +616,96 @@ void cgroup_config_cleanup_namespace_table(void)
 	memset(&config_namespace_table, 0, sizeof(struct cg_mount_table_s) * CG_CONTROLLER_MAX);
 }
 
+/**
```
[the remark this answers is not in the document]
kamalesh-babulal: Agreed, will remove them.

src/config.c
```diff
@@ -887,6 +985,49 @@ static int config_validate_namespaces(void)
...
+static int config_create_slice_scope(char * const _cgroup_systemd_default_slice_name)
+{
+	struct cgroup_systemd_scope_opts scope_opts;
+	struct cgroup_systemd_opts *curr;
+	int ret = 0;
+
+	if (!_cgroup_systemd_default_slice_name)
+		return 1;
+
+	scope_opts.delegated = 1;
```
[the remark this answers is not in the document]
kamalesh-babulal: :) Yes, I should be calling it instead.

src/config.c
```diff
@@ -608,6 +616,96 @@ void cgroup_config_cleanup_namespace_table(void)
...
+ * Returns the ret value from cgroup_add_systemd_opts() and 0 on failure, also
+ * free() the allocated objects.
+ */
+int cgroup_alloc_systemd_opts(const char * const config, const char * const value)
+{
+	struct cgroup_systemd_opts *new;
+
+	if (strncmp(config, "slice\0", 6)   != 0 &&
+	    strncmp(config, "scope\0", 6)   != 0 &&
+	    strncmp(config, "setdefault\0", 11) != 0)
```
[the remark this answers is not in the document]
drakenclimber: Thinking out loud here:

What if we did something like this?  Then the user could pick any arbitrary relative path.  Thoughts?
```
systemd {
        slice = testing;
        scope = testing;
}

default_cgroup_path {
    testing.slice/testing.scope
}
```

include/libcgroup/config.h
```diff
@@ -29,13 +29,57 @@ extern "C" {
  * and applications can benefit from them.
  */
 
+/*
+ * cgroup systemd settings.
+ */
+struct cgroup_systemd_opts {
```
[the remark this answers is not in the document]
kamalesh-babulal: Given that, `systemd` will be heavily used and it will be enabled by default in the configurable option, we can very well remove it, on the other hand, we would wait for the feedback from the community and remove it from the release + 1 having configurable support for systemd, after fixing any bugs reported.

src/config.c
```diff
@@ -1061,6 +1204,39 @@ static int cgroup_parse_config(const char *pathname)
 	return ret;
 }
 
+/**
+ * Parse the systemd delegation settings from the configuration file
+ * and get the default delegation cgroup name, if set.
+ *
+ * @param pathname Name of the configuration file to load.
+ */
+int cgroup_get_default_slice_name(const char * const pathname)
```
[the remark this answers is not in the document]
kamalesh-babulal: Agreed, should it be renamed to `cgroup_set_default_slice_name` or another other better name?

src/config.c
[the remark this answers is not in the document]
drakenclimber: And I don't really love the `default_cgroup_path` name, so I'm open to ideas there.

tests/Makefile.am
kamalesh-babulal: My bad, was testing some bug with Gunit.

src/config.c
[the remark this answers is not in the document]
drakenclimber: That sounds good to me

src/tools/cgget.c
kamalesh-babulal: @drakenclimber Thanks for all the corner cases and functionality, I had missed. I will re-work them and share the changes sooner.

kamalesh-babulal:
**V2 Changes**
- renamed `cgroup_get_default_slice_name()` to `cgroup_set_default_systemd_cgroup()`
- Improved the naming of the systemd configuration examples
- The default delegated cgroup path is `<slice name>.slice/<scope name>.scope`, whereas in the previous version it was just   `<slice name>.slice/`
- Introduced `WITH_SYSTEMD CFLAGS` for `src/*` and `src/tools/*` that would guard for systemd code  across the sources
- Introduced `/var/run/libcgroug/systemd` to store the default delegated slice/scope name, this addresses the issue of multiple` cgconfig.conf` files passed via `-d` option to the `cgconfigparser`.
- Introduced `systemd_default_cgroup_lock` to avoid race conditions on file operation on `systemd_default_cgroup_file` (`/var/run/libcgroup/systemd)`
- `cg_build_path_locked()` - redid the logic cgroup name starting with a '/'  where the '/<cgroup name> and '/' are honored while constructing the path using the systemd_default_cgroup.
- `cg_build_path_locked()` - fix the `snprintf` warning using a temporary variable to construct the path.
- `tools/*` will no longer pass the CGCONFIG_CONF_FILE, instead read from the  `systemd_default_cgroup_file` (`/var/run/libcgroup/systemd`)
- `tools/*` will only display options of overriding the `systemd_default_cgroup`,  as well as constructing a path with `systemd_default_cgroup_file` iff configured using systemd.
- Other improvements in error checkings and added more comments.

drakenclimber (commented):
I think this is looking pretty good.  Seems like we're going to need a ton of tests to verify this doesn't break all the various cgget/cgset options.  There are already a lot of permutations there, and this adds even more :)

include/libcgroup/config.h
```diff
@@ -29,13 +29,56 @@ extern "C" {
  * and applications can benefit from them.
  */
 
+/*
```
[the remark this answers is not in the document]
kamalesh-babulal: @drakenclimber Thanks a lot for the review. Sure, make sense to move them.

src/config.c
```diff
@@ -887,6 +908,7 @@ static int config_validate_namespaces(void)
 	return error;
 }
 
+
```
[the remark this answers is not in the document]
kamalesh-babulal: Will fix it in the next version.

src/api.c
```diff
@@ -1684,8 +1684,36 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
...
+		goto out;
+	}
+
+#ifdef WITH_SYSTEMD
+	/*
+	 * If the user specifies the name as /<cgroup-name>, they are
+	 * effectively overriding the systemd_default_cgroup.  If the name
+	 * is '/', path is cgroup root or systemd_default_cgroup root.
+	 */
+	if (strlen(systemd_default_cgroup) && name && name[0] == '/' && name[1] != '\0')
```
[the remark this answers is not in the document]
kamalesh-babulal: There are chances the user doesn't set the `systemd->setdefault` in the configuration, hence `systemd_default_cgroup` can be zero length.  override it only if its set, else `tmp_systemd_default_cgroup` will be zero-length string.

src/api.c
```diff
@@ -1684,8 +1684,36 @@ char *cg_build_path_locked(const char *name, char *path, const char *type)
...
+		tmp_systemd_default_cgroup[0] = '\0';
+	else
+		snprintf(tmp_systemd_default_cgroup, len, "%s/", (systemd_default_cgroup));
```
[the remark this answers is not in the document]
kamalesh-babulal: The extra byte is required for the `/`, to keep GCC happy `snprintf()` space allocation. The global `systemd_default_group` has only one `/` (slice/scope)

src/config.c
```diff
@@ -2044,6 +2044,87 @@ static int config_create_slice_scope(char * const tmp_systemd_default_cgroup)
...
+	case CGROUP_MODE_UNK:
+		break;
+	}
+
+	if (!strlen(path)) {
+		cgroup_dbg("Unable to form canonical path for %s", systemd_default_cgroup);
+		goto err;
+	}
+
+	snprintf(tmp, sizeof(tmp), "%s/cpu.stat", path);
```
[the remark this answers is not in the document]
kamalesh-babulal: Agree, using `cgroup.procs` is reliable over `cpu.stat`.

src/config.c
```diff
@@ -2044,6 +2044,87 @@ static int config_create_slice_scope(char * const tmp_systemd_default_cgroup)
...
+	if (!systemd_def_cgrp_f) {
+		cgroup_dbg("Unable to read %s ", systemd_default_cgroup_file);
+		goto err;
+	}
+
+	len = fread(systemd_default_cgroup, sizeof(char), FILENAME_MAX, systemd_def_cgrp_f);
+	fclose(systemd_def_cgrp_f);
+	/* at the minimum it should be <n>.slice/<n>.scope == 15 */
+	if (len < 15) {
+		cgroup_dbg("Malformed system default cgroup %s", systemd_default_cgroup);
```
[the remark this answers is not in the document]
kamalesh-babulal: Will fix it in the next iteration.

src/tools/cgget.c
```diff
@@ -583,6 +599,19 @@ static int fill_empty_controller(struct cgroup * const cg, struct cgroup_control
...
 
 	path_len = strlen(path);
+#ifdef WITH_SYSTEMD
+	if (cg->name[0] == '/' && strncmp(path + (path_len - 7), ".scope/", 7) == 0) {
+		snprintf(tmp, FILENAME_MAX, "%s", dirname(path));
+		strncpy(path, tmp, FILENAME_MAX - 1);
+		path[FILENAME_MAX - 1] = '\0';
+	}
+	path_len = strlen(path);
+	if (cg->name[0] == '/' && strncmp(path + (path_len - 6), ".slice/", 6) == 0) {
```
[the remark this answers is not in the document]
kamalesh-babulal: The first comparison is made for the `7` bytes, because the path ends with a `/` and that needs to be taken into account  but after trimming the scope cgroup from the path using `dirname()`, which also deletes the `/`, hence the second comparison is done for 6 characters but you are right, that the comparison is misleading and `.slice` should be used instead.

src/tools/cgget.c
```diff
@@ -583,6 +599,19 @@ static int fill_empty_controller(struct cgroup * const cg, struct cgroup_control
 		goto out;
 
 	path_len = strlen(path);
+#ifdef WITH_SYSTEMD
+	if (cg->name[0] == '/' && strncmp(path + (path_len - 7), ".scope/", 7) == 0) {
```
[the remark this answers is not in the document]
kamalesh-babulal: Sorry, will add comments here.

src/api.c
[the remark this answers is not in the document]
drakenclimber: Sounds like a really good thing to add a test around :)

kamalesh-babulal:
**V3 Changes**
Addressed the review  comments from Tom  and a few notable changes are:
- Moved all the definitions to `systemd.h`.
- Use `cgroup.procs` instead of  `cpu.stat` to check if the cgroup exists.
- Added documentation at a couple of places.
- Introduce `cgget` changes to `cgxget` too.

drakenclimber:
I think this is looking pretty good.  Let's get some tests around it and see what they find.

A while back I prototyped a couple commits that help with testing cgget in this patchset.  They may be a good start:
https://github.com/drakenclimber/libcgroup/commit/7bdb5f14926bea396ec0b344a548e274c06218a4
https://github.com/drakenclimber/libcgroup/commit/2afe401bcf324a6c07fda269b03dd601297c6476

It looks like I didn't fully flesh out the `ignore_systemd` flag in the first patch, but it's close.

src/api.c
```diff
@@ -1674,7 +1679,51 @@ static char *cg_concat_path(const char *pref, const char *suf, char *path)
...
+	 * multiple cgroups, hence use temporary variable for manipulations
+	 * for example:
+	 * cgget -g cpu:/ -g cpu:cgrp1 -g cpu:/cgrp2
+	 */
+	tmp_systemd_default_cgroup = malloc(sizeof(char) * len);
+	if (!tmp_systemd_default_cgroup) {
+		cgroup_err("Failed to allocate memory for tmp_systemd_default_cgroup.\n");
+		goto out;
+	}
+	memset(tmp_systemd_default_cgroup, 0 , sizeof(*tmp_systemd_default_cgroup));
```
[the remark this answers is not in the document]
kamalesh-babulal: @Werkov Thanks for the review. Sure, will replace it using `calloc()`

src/config.c
```diff
@@ -1807,3 +1855,287 @@ int cgroup_config_create_template_group(struct cgroup *cgroup, char *template_na
 	cgroup_free(&aux_cgroup);
 	return ret;
 }
+
+#ifdef WITH_SYSTEMD
+int cgroup_add_systemd_opts(const char * const config, const char * const value)
+{
+	struct cgroup_systemd_opts *curr = cgroup_systemd_opts_tail;
+
+	if (strncmp(config, "slice\0", 6) == 0)
```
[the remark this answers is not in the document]
kamalesh-babulal: I wanted to use `strncmp()` to limit the check to `slice`, given that it's read from the configuration file, the user can always give a wrong/lengthier configuration option.  On the other hand, agree that it's unlikely, will change these constructions with `strcmp()`.

kamalesh-babulal:
**V4 Changes**:
- Addressed the review comments from Michal Koutný
- Added systemd config option `pid` that lets the users add an existing task as the scope task in place of the `idle thread`
- Added test cases from Tom Hromatka
- Added test case to test `pid` based on the test cases written by Tom Hromatka

src/tools/cgset.c
```diff
@@ -139,8 +144,16 @@ int main(int argc, char *argv[])
...
 	/* parse arguments */
-	while ((c = getopt_long (argc, argv, "r:h", long_options, NULL)) != -1) {
+#ifdef WITH_SYSTEMD
+	while ((c = getopt_long (argc, argv, "r:hb", long_options, NULL)) != -1) {
 		switch (c) {
+		case 'b':
+			ignore_default_systemd_delegate_slice = 1;
+			break;
+#else
+	while ((c = getopt_long (argc, argv, "r:hb", long_options, NULL)) != -1) {
```
drakenclimber, line 154: Typo - should be `r:h`

src/tools/cgdelete.c
```diff
@@ -150,9 +154,16 @@ int main(int argc, char *argv[])
...
-	while ((c = getopt_long(argc, argv, "rhg:",
-		long_options, NULL)) > 0) {
+#ifdef WITH_SYSTEMD
+	while ((c = getopt_long(argc, argv, "rhg:b", long_options, NULL)) > 0) {
 		switch (c) {
+		case 'b':
+			ignore_default_systemd_delegate_slice = 1;
+			break;
+#else
+	while ((c = getopt_long(argc, argv, "rhg:b", long_options, NULL)) > 0) {
```
drakenclimber, line 164: typo - should be `rhg:`

src/tools/cgexec.c
```diff
@@ -63,9 +67,16 @@ int main(int argc, char *argv[])
 	int c;
 
 	memset(cgroup_list, 0, sizeof(cgroup_list));
-
+#ifdef WITH_SYSTEMD
 	while ((c = getopt_long(argc, argv, "+g:sh", longopts, NULL)) > 0) {
```
drakenclimber, line 71: I think this should be `+g:shb`, correct?

tests/ftests/054-sudo-g_flag_controller_only_systemd-v1.py
```diff
@@ -0,0 +1,158 @@
...
+        cause = 'This test cannot be run within a container'
+
+    return result, cause
+
+
+def setup(config):
+    # Let's remove the existing default systemd cgroup file
+    # so that the temp cgroup we create is at the know path.
+    if os.path.exists(DEFAULT_SYSTEMD_CGROUP_NAME):
+        os.remove(DEFAULT_SYSTEMD_CGROUP_NAME)
```
drakenclimber, line 51: This feels like belt and suspenders code to me - which isn't a bad thing.

But, it feels weird in this test.  Is a user expected to do this?  Shouldn't calling `configparser()` automatically cause this to be overwritten.  And if `configparser()` should overwrite this, are we potentially hiding a bug because we aren't doing things the same way a user would?

tests/ftests/054-sudo-g_flag_controller_only_systemd-v1.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # The job of the tmp cgroup is to create a task, that will be
+    # used by the systemd as pid for the scope.
+    Cgroup.create(config, CONTROLLER, TMP_CGNAME)
+    config.process.create_process_in_cgroup(config, CONTROLLER, TMP_CGNAME)
```
drakenclimber, line 56: Why are you creating a process in a cgroup only to have it be moved to the newly-created scope?  Wouldn't calling `config.create_process()` be simpler?

tests/ftests/054-sudo-g_flag_controller_only_systemd-v1.py
kamalesh-babulal: Agreed with all the points, let me remove them so that when fails we  know  something  went wrong, which is the purpose of the test :)

src/tools/cgset.c
kamalesh-babulal: :( sorry for typos will fix them

tests/ftests/054-sudo-g_flag_controller_only_systemd-v1.py
kamalesh-babulal: yeah, it will be way more elegant.

src/config.c
```diff
@@ -91,6 +92,26 @@ static struct cgroup_string_list *template_files;
...
+static const char * const systemd_default_cgroup_file = "/var/run/libcgroup/systemd";
+
+/* Lock that protects systemd_default_cgroup_file */
+static pthread_rwlock_t systemd_default_cgroup_lock = PTHREAD_RWLOCK_INITIALIZER;
+
+/*
+ * cgroup_systemd_opts is dynamically allocated, hence having head
+ * and tail helps in traversing.
+ */
+static struct cgroup_systemd_opts *cgroup_systemd_opts_head;
```
drakenclimber, line 109: Unless I'm missing something, I think we need to initialize these to NULL.

tests/ftests/systemd.py
```diff
@@ -27,3 +29,91 @@ def is_delegated(config, scope_name):
...
+        if ctrl_name is None:
+            if ignore_systemd:
+                raise IndexError('Unsupported controller \'None\'')
+
+            controllers = ['cpu', 'memory', 'io', 'pids', 'cpuset']
+            for controller in controllers:
+                ctrl_mnt = Cgroup.get_controller_mount_point(controller)
+                if 'Unknown version' not in ctrl_mnt:
+                    break
+            else:
```
drakenclimber, line 50: nifty use of the `for/else` loop :)

tests/ftests/systemd.py
```diff
@@ -27,3 +29,91 @@ def is_delegated(config, scope_name):
...
+        # on cgroup v2 is unified setup. But it doesn't have a mount point,
+        # lets use the first available controller mount point to form the path
+        if ctrl_name is None:
+            if ignore_systemd:
+                raise IndexError('Unsupported controller \'None\'')
+
+            controllers = ['cpu', 'memory', 'io', 'pids', 'cpuset']
+            for controller in controllers:
+                ctrl_mnt = Cgroup.get_controller_mount_point(controller)
+                if 'Unknown version' not in ctrl_mnt:
```
drakenclimber, line 48: I don't think this `if` is buying you anything.  Unknown version and unknown mount point are raised as Exceptions, and thus I don't think this `if` is protecting you.

tests/ftests/systemd.py
```diff
@@ -27,3 +29,91 @@ def is_delegated(config, scope_name):
...
+    @staticmethod
+    def cgroup_exists(config, ctrl_name, cgroup_name, ignore_systemd=False):
+        # 'None' is a valid controller name, while creating an empty cgroup
+        # on cgroup v2 is unified setup. But it doesn't have a mount point,
+        # lets use the first available controller mount point to form the path
+        if ctrl_name is None:
+            if ignore_systemd:
+                raise IndexError('Unsupported controller \'None\'')
+
+            controllers = ['cpu', 'memory', 'io', 'pids', 'cpuset']
```
drakenclimber, line 45: Does it make sense to have a list of controllers here, or does it make more sense to make `get_controller_mount_point()` smarter and be able to handle a `None` controller?

tests/ftests/060-sudo-cgconfigparser-systemd.py
```diff
@@ -0,0 +1,194 @@
...
+        try:
+            Cgroup.configparser(config, load_file=CONFIG_FILE_NAME)
+        except RunError as re:
+            if configuration[1] not in re.stdout:
+                result = consts.TEST_FAILED
+                tmp_cause = (
+                            'Unexpected error {}, while parsing configuration:'
+                            '\n{}'.format(re.stdout, configuration[0])
+                        )
+                cause = '\n'.join(filter(None, [cause, tmp_cause]))
```
drakenclimber, line 107: This is a cool use of `filter()`.  Nice :)

tests/ftests/065-sudo-systemd_cgclassify-v1.py
```diff
@@ -0,0 +1,196 @@
...
+    except RunError as re:
+        err_str = 'Error changing group of pid {}: Cgroup does not exist'.format(OTHER_PIDS)
+        if re.stderr != err_str:
+            raise re
+    else:
+        result = consts.TEST_FAILED
+        cause = 'Changing group of pid {} erroneously succeeded'.format(OTHER_PIDS)
+
+    # classify the task from the non-systemd scope cgroup to systemd scope
+    # cgroup (should fail due to wrong path).
```
drakenclimber, line 135: nitpick - I think this comment isn't right.

tests/ftests/066-sudo-systemd_cgclassify-v2.py
```diff
@@ -0,0 +1,213 @@
...
+    except RunError as re:
+        err_str = 'Error changing group of pid {}: No such file or directory'.format(OTHER_PIDS)
+        if re.stderr != err_str:
+            raise re
+    else:
+        result = consts.TEST_FAILED
+        cause = 'Changing group of pid {} erroneously succeeded'.format(OTHER_PIDS)
+
+    # classify the task from the non-systemd scope cgroup to systemd scope
+    # cgroup (should fail due to wrong cgroup path).
```
drakenclimber, line 152: nitpick - same comment as on the previous test.  I think this comment is a copy/paste error :)

tests/ftests/067-sudo-systemd_cgexec-v1.py
```diff
@@ -0,0 +1,199 @@
...
+        result = consts.TEST_FAILED
+        cause = (
+                    'Failed to create cgroup {} under '
+                    '/sys/fs/cgroup/{}/'.format(OTHER_CGNAME, CONTROLLER)
+                )
+
+    return result, cause
+
+
+def create_process_get_pid(config, CGNAME, SLICE='', ignore_systemd=False):
```
drakenclimber, line 77: nitpick - you probably don't want to override the global variable named `SLICE`.  It'll work as you hope, but it's kinda a maintenance nightmare

tests/ftests/067-sudo-systemd_cgexec-v1.py
```diff
@@ -0,0 +1,199 @@
...
+        return result, cause
+
+    OTHER_PIDS, result, tmp_cause = create_process_get_pid(
+                                                                config, OTHER_CGNAME,
+                                                                ignore_systemd=True
+                                                          )
+    if result == consts.TEST_FAILED:
+        return result, cause
+
+    # classify the task from the non-systemd scope cgroup to systemd scope
```
drakenclimber, line 125: I don't understand the need to classify the processes to another cgroup.  Isn't the point of this test to run cgexec?  If so, wouldn't a clearer validation simply be to verify that the process(es) were created in the requested cgroups?

tests/ftests/068-sudo-systemd_cgexec-v2.py
```diff
@@ -0,0 +1,216 @@
...
+        result = consts.TEST_FAILED
+        cause = (
+                    'Failed to create cgroup {} under '
+                    '/sys/fs/cgroup/{}/'.format(OTHER_CGNAME, CONTROLLER)
+                )
+
+    return result, cause
+
+
+def create_process_get_pid(config, CGNAME, SLICE='', ignore_systemd=False):
```
drakenclimber, line 89: same nitpick comment as previous test

tests/ftests/069-sudo-systemd_cgxget-cpu-settings-v1.py
```diff
@@ -0,0 +1,169 @@
...
+
+def test(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    cgrps = {SYSTEMD_CGNAME: False, OTHER_CGNAME: True}
+    for i in cgrps:
+        for entry in TABLE:
+            Cgroup.xset(config, cgname=i, setting=entry[0], value=entry[1],
+                        version=entry[2], ignore_systemd=cgrps[i])
```
drakenclimber, line 117: Dictionaries are only ordered on Python 3.7 and above.  Do we care that this code won't work on older versions of Python?  Should we enforce that?

src/config.c
```diff
@@ -1807,3 +1857,236 @@ int cgroup_config_create_template_group(struct cgroup *cgroup, char *template_na
...
+{
+	struct cgroup_systemd_opts *new_cgrp_systemd_opts;
+
+	/*
+	 * check for the allowed systemd configurations. We don't
+	 * check for values, they will be checked by systemd anyway.
+	 */
+	if (strcmp(config, "slice") != 0 &&
+	    strcmp(config, "scope") != 0 &&
+	    strcmp(config, "setdefault") != 0 &&
```
drakenclimber, line 1911: My apologies up front.... :)

Now that I think about it more, would naming this parameter `default` (rather than `setdefault`) make more sense?  We didn't name the other parameters `setslice`, `setscope`, etc. but they're obviously setting something.

src/tools/lscgroup.c
```diff
@@ -47,6 +47,9 @@ static void usage(int status, const char *program_name)
 	info("  -g <controllers>:<path>	Control group to be ");
```
drakenclimber, line 47: nitpick - First line of the commit comment mentions `cgcreate` :)

src/tools/cgclassify.c
```diff
@@ -41,6 +41,9 @@ static void usage(int status, const char *program_name)
 	info("  --cancel-sticky		cgred daemon change pidlist and children tasks\n");
```
drakenclimber, line 41: nitpick - `cgcreate` is mentioned twice in the commit comment :)

src/tools/cgexec.c
```diff
@@ -49,11 +49,15 @@ static void usage(int status, const char *program_name)
 	info("  -h, --help			Display this help\n");
```
drakenclimber, line 49: nitpick - first line of the commit comment mentions `cgcreate` :)

tests/ftests/systemd.py
```diff
@@ -27,3 +29,91 @@ def is_delegated(config, scope_name):
                 # until we figure out something better :(
                 return True
             raise re
+
+    # This function builds the absolute path of the cgroup, using the default
+    # systemd path and checks if the cgroup exists. The function can build
+    # path for all three cgroup setup modes, including empty cgroup v2.
+    @staticmethod
+    def cgroup_exists(config, ctrl_name, cgroup_name, ignore_systemd=False):
```
drakenclimber, line 37: Hmmm, thinking out loud here...  Currently you aren't using `config` but would we need it if this function were run within a container?  I think so

tests/ftests/systemd.py
```diff
@@ -27,3 +29,91 @@ def is_delegated(config, scope_name):
...
+            return False
+
+        cgrp_path = os.path.join(ctrl_mnt, slice_scope_path, cgroup_name)
+        return os.path.isdir(cgrp_path)
+
+    # This is a simple wrapper, that calls Cgroup.create() to create the
+    # cgroup and cgroup_exists(), which builds the cgroup path and returns
+    # True if the cgroup exists, False otherwise
+    @staticmethod
+    def create_and_validate(config, ctrl_name, cgroup_name, ignore_systemd=False):
```
drakenclimber, line 72: `create()` is located in the Cgroup class.  Would it make sense to have `create_and_validate()` next to it in the same class?  (It seems like we should consider converting some of the `create()` calls in other tests to `create_and_validate()` and thus putting this in Cgroup class might make more sense.

tests/ftests/systemd.py
drakenclimber, line 37: Also, would it make more sense to have this function in the Cgroup class, and it could just be called `Cgroup.exists()`?

tests/ftests/069-sudo-systemd_cgxget-cpu-settings-v1.py
drakenclimber: I'm specifically referring to the `cgrps[i]` snippet of code

src/config.c
kamalesh-babulal: @drakenclimber Thank you so much for the review.
Isn't `static` storage initialized to a null pointer unless initialized explicitly?
```
$ echo -e "struct { int num;}a;\n struct a *foo;\n int main() {\n if (foo) { return 1;}\n return 0;\n}" | gcc -x c - -o -
$ echo $?
0
```

src/config.c
kamalesh-babulal: I completely agree, in fact, I had thought about renaming it to `default` will rename it.

kamalesh-babulal: actually, `default` can't be used as its a keyword, does `delegate` sound better?

src/tools/lscgroup.c
kamalesh-babulal: copy, and paste typos, I tried being cautious by still, sorry :(

tests/ftests/systemd.py
kamalesh-babulal: @drakenclimber smart :) Let me implement it

kamalesh-babulal: Something like below:
```
--- a/tests/ftests/cgroup.py
+++ b/tests/ftests/cgroup.py
@@ -741,6 +741,10 @@ class Cgroup(object):
             for line in mntf.readlines():
                 mnt_path = line.split()[1]
 
+                if ctrl_name is None:
+                    if line.split()[0] == 'cgroup2':
+                        return mnt_path
+
                 if line.split()[0] == 'cgroup':
                     for option in line.split()[3].split(','):
                         if option == ctrl_name:
@@ -758,6 +762,9 @@ class Cgroup(object):
                 mnt_path = line.split()[1]
 
                 if line.split()[0] == 'cgroup2':
+                    if ctrl_name is None:
+                        return mnt_path
+
                     ctrl_file = os.path.join(line.split()[1],
                                              'cgroup.controllers')
 
@@ -774,7 +781,16 @@ class Cgroup(object):
 
     @staticmethod
     def get_controller_mount_point(ctrl_name):
-        vers = CgroupVersion.get_version(ctrl_name)
+        if ctrl_name is None:
+            mode = int(Cgroup.get_cgroup_mode(None))
+            # map return modes:
+            # CGROUP_MODE_LEGACY/CGROUP_MODE_HYBRID ->  CgroupVersion.CGROUP_V1
+            # CGROUP_MODE_UNIFIED -> CgroupVersion.CGROUP_V2
+            if mode >= 2:
+                mode = mode - 1
+            vers = CgroupVersion(mode)
+        else:
+            vers = CgroupVersion.get_version(ctrl_name)
 
         if vers == CgroupVersion.CGROUP_V1:
             return Cgroup.__get_controller_mount_point_v1(ctrl_name)
```

tests/ftests/systemd.py
kamalesh-babulal: Something like this with `get_controller_mount_point()` support for `None`
```
+    @staticmethod
+    def exists(config, ctrl_name, cgroup_name, ignore_systemd=False):
+        ctrl_mnt = Cgroup.get_controller_mount_point(ctrl_name)
+
+        if (ignore_systemd):
+            cgrp_path = os.path.join(ctrl_mnt, cgroup_name)
+        else:
+            try:
+                cmd = [ 'cat', '/run/libcgroup/systemd']
+                if config.args.container:
+                    slice_scope_path = config.container.run(cmd, shell_bool=True)
+                else:
+                    slice_scope_path = Run.run(cmd, shell_bool=True)
+            except RunError:
+                    return False
+
+            cgrp_path = os.path.join(ctrl_mnt, slice_scope_path, cgroup_name)
+
+        proc_file =  os.path.join(cgrp_path, 'cgroup.procs')
+        cmd = ['cat', proc_file]
+        if config.args.container:
+            try:
+                exists = config.container.run(cmd, shell_bool=True)
+            except RunError:
+                return False
+        else:
+            try:
+                exists = Run.run(cmd, shell_bool=True)
+            except RunError:
+                return False
+
+        return True
```

kamalesh-babulal: Sure, that will make it more `bash` than `Python` :)

tests/ftests/065-sudo-systemd_cgclassify-v1.py
kamalesh-babulal: Does the following re-phrasing make sense?
```      
# This should fail because of trying to migrate a task to SYSTEMD_CGNAME
# under the wrong path from OTHER_CGNAME.
```

tests/ftests/067-sudo-systemd_cgexec-v1.py
kamalesh-babulal: Agree, will rename it `SLICENAME` to avoid confusion.

tests/ftests/067-sudo-systemd_cgexec-v1.py
kamalesh-babulal: Agreed, just checking for process creation under the right cgroup should be sufficient and the migration test is not required here

tests/ftests/069-sudo-systemd_cgxget-cpu-settings-v1.py
kamalesh-babulal: The unordered Dictionaries will not harm there, because `cgrps[SYSTEMD_CGNAME] = > ignore_systemd=False` and `cgrps[OTHER_CGNAME] => ignore_systemd=True` always no matter the order, and `xset` will find the right path of the cgroup.

tests/ftests/065-sudo-systemd_cgclassify-v1.py
drakenclimber: It's a copy-paste of the previous comment.  Would this be correct?
```
# classify the task from the systemd scope cgroup to the non-systemd scope
# cgroup (should fail due to wrong path).
```

tests/ftests/069-sudo-systemd_cgxget-cpu-settings-v1.py
drakenclimber: Thanks.  Makes sense

src/config.c
drakenclimber: Argh... thanks for digging.  Let's stick with `setdefault` then.  `delegate` is likely an option we will need to add later, but it should apply to the `opts->delegate` field within the systemd options structure.

tests/ftests/systemd.py
drakenclimber: I don't have a strong preference here.  If you haven't changed it, leave it as is.  This is internal to the tests and we can change its parameters later with no repercussions.

tests/ftests/065-sudo-systemd_cgclassify-v1.py
kamalesh-babulal: Agreed, re-reading it, it's not obvious. Does the more verbose version make sense?
```
    # classify a task from the non-systemd scope cgroup (OTHER_CGNAME) to systemd                   
    # scope cgroup (SYSTEMD_CGNAME).  Migration should fail due to the incorrect                    
    # destination cgroup path that gets constructed, without the systemd slice/scope                
    # when ignore_systemd is set) 
```

src/config.c
kamalesh-babulal: Ah, Make sense, will leave it unchanged.

tests/ftests/systemd.py
kamalesh-babulal: Sure, I have re-written it :)

kamalesh-babulal:
### V6 Changes:
- Addressed the review comments
- Included the newly added test cases to the `tests/ftests/Makefile.am`

drakenclimber:
Pushed to main.  Thanks so much @kamalesh-babulal for all the hard work and time!

I'll merge it into release-3.0 in a bit.

drakenclimber:
Pushed to release-3.0.  Let me know if you notice anything awry.  Thanks, all!

PR #250, opened 2022-09-16
Add initial systemd support

drakenclimber:
Add a C API, `cgroup_create_scope()`, and the associated python binding, `Cgroup.create_scope()`, to create a cgroup that's a valid systemd scope.  If the user has requested it to be delegated, then systemd will relinquish management of this cgroup and its children.

Future pull requests will add hooks into the existing C APIs, CLI tools, etc. as necessary.

PR #270, opened 2022-11-18
Add a libcgroup-style function to create systemd scopes

drakenclimber:
This patchset builds off of the first systemd pull request #250 that was merged a couple weeks ago.  It adds a function, `cgroup_create_scope2()`, that will create a systemd scope using libcgroup-style parameters like `struct cgroup *cgroup`.  (I'm open to other names for the function if anyone has a good suggestion.)

I think this should add a pretty good bump to code coverage as well because I am now testing functions that set the permissions and owner/group of a cgroup.  [Here's a link](https://coveralls.io/builds/54366619) to the code coverage run on my personal github.


src/systemd.c
```diff
@@ -281,3 +283,50 @@ int cgroup_create_scope(const char * const scope_name, const char * const slice_
...
+		goto err;
+
+	/*
+	 * cgroup_create_cgroup() can gracefully handle EEXIST if the cgroup already exists, so
+	 * let's just call it because it already knows how to manage the ownership of the cgroup
+	 * hierarchy.
+	 */
+	error = cgroup_create_cgroup(cgroup, ignore_ownership);
+	if (error)
+		goto err;
```
kamalesh-babulal, line 323: @drakenclimber Thanks for working on the idea, it looks awesome.  Can you please add more comments on why we need to call `cgroup_create_cgroup()`?

src/python/libcgroup.pyx
```diff
@@ -372,6 +372,42 @@ cdef class Cgroup:
...
+        tasks_gid - gid that should own the tasks file
+        ctrl_uid - uid to recursively apply to the entire cgroup hierarchy
+        ctrl_gid - gid to recursively apply to the entire cgroup hierarchy
+
+        Note:
+        Does not modify the cgroup sysfs.  Does not read from the cgroup sysfs
+        """
+        ret = cgroup.cgroup_set_uid_gid(self._cgp, tasks_uid, tasks_gid, ctrl_uid, ctrl_gid)
+        if ret is not 0:
+            raise RuntimeError("cgroup_create_scope failed: {}".format(ret))
```
kamalesh-babulal, line 389: `RuntimeError` has wrong error message :)

src/systemd.c
kamalesh-babulal: I guess the check for `error` can be removed too.

src/python/libcgroup.pyx
```diff
@@ -372,6 +372,42 @@ cdef class Cgroup:
...
+        Arguments:
+        dir_mode - permissions to set on the cgroup directory
+        ctrl_mode - permissions to set on the files in the directory, except tasks
+        task_mode - permissions to set on the tasks file
+
+        Note:
+        Does not modify the cgroup sysfs.  Does not read from the cgroup sysfs.  Use the stat
+        module to populate the modes [1]
+
+        [1] https://docs.python.org/3/library/stat.html
```
kamalesh-babulal, line 403: Is the comment about populate modes is not clear.  Is that intended for default values?

tests/ftests/051-sudo-set_uid_gid_v2.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: LGPL-2.1-only
+#
+# Test to set the uid/gid on a cgroup using the python bindings
```
kamalesh-babulal, line 4: Do you think adding `Cgroup v2` to the test description, might help. Should the copyright year also need to change?

tests/ftests/051-sudo-set_uid_gid_v2.py
```diff
@@ -0,0 +1,90 @@
...
+    cg.create(ignore_ownership=False)
+
+    ctrl_path = os.path.join(CgroupCli.get_controller_mount_point(CONTROLLER), CGNAME,
+                             'cgroup.procs')
+
+    uid = utils.get_file_owner_uid(config, ctrl_path)
+    if uid != CTRL_UID:
+        result = consts.TEST_FAILED
+        cause = 'Expected cgroup.procs owner to be {} but it\'s {}'.format(CTRL_UID, uid)
+
```
kamalesh-babulal, line 57: Is the fallthrough intentional? If both `uid` and `gid` checks fails,  then the `gid`  will override the `uid cause`

tests/ftests/052-sudo-set_uid_gid_v1.py
```diff
@@ -0,0 +1,103 @@
...
+    cg.add_controller(CONTROLLER)
+    cg.create(ignore_ownership=False)
+
+    tasks_path = os.path.join(CgroupCli.get_controller_mount_point(CONTROLLER), CGNAME,
+                              'tasks')
+
+    tasks_uid = utils.get_file_owner_uid(config, tasks_path)
+    if tasks_uid != TASKS_UID:
+        result = consts.TEST_FAILED
+        cause = 'Expected tasks owner to be {} but it\'s {}'.format(TASKS_UID, tasks_uid)
```
kamalesh-babulal, line 56: Fallthrough here, will override the causes incase two or more checks fails.

tests/ftests/052-sudo-set_uid_gid_v1.py
```diff
@@ -0,0 +1,103 @@
...
+        result = consts.TEST_FAILED
+        cause = 'Expected cgroup.procs group to be {} but it\'s {}'.format(CTRL_GID, ctrl_gid)
+
+    return result, cause
+
+
+def teardown(config):
+    CgroupCli.delete(config, CONTROLLER, CGNAME)
+
+    return consts.TEST_PASSED, None
```
kamalesh-babulal, line 82: this return is not used in `main` and the return is True always, should this be removed?

tests/ftests/051-sudo-set_uid_gid_v2.py
```diff
@@ -0,0 +1,90 @@
...
+        result = consts.TEST_FAILED
+        cause = 'Expected cgroup.procs group to be {} but it\'s {}'.format(CTRL_GID, gid)
+
+    return result, cause
+
+
+def teardown(config):
+    CgroupCli.delete(config, None, CGNAME)
+
+    return consts.TEST_PASSED, None
```
kamalesh-babulal, line 69: this return is not used in main and the return is True always, should this be removed?

src/python/libcgroup.pyx
kamalesh-babulal: I understand the comment now, after the test cases.

src/systemd.c
drakenclimber: Good idea on both accounts.

src/python/libcgroup.pyx
drakenclimber: D'oh :(

tests/ftests/051-sudo-set_uid_gid_v2.py
drakenclimber: Good idea on both accounts.

tests/ftests/051-sudo-set_uid_gid_v2.py
drakenclimber: Honestly I didn't really think it through too hard.  I made sure that success couldn't accidentally overwrite a failure, but I didn't worry too much about a failure overwriting another failure.  I'll see if I can concatenate the failures into a single `cause`

tests/ftests/052-sudo-set_uid_gid_v1.py
drakenclimber: Good find.  I have a proposed fix; let me know your thoughts

kamalesh-babulal: I like the idea of concatenation, it is clean and clever.

kamalesh-babulal (approved):

drakenclimber:
Conflicts were trivial.  Changes have been pushed to main and release-3.0.  Thanks for the review, @kamalesh-babulal.

PR #274, opened 2022-12-07
workflows/ci: use ubuntu 20.04 for cgroup v1 functional test [v2.0]

kamalesh-babulal:
Using ubuntu-latest (Ubuntu 22.04) for the cgroup v1 functional test fails, due to two reasons:
- The lxd package has been renamed to lxd-installer
- Ubuntu-latest (22.04) supports only cgroup v2 hierarchy and for branch-2.0 pass the test cases, requires forward porting a couple of patches from 3.0 branch. Let's run it on the known stable release Ubuntu-20.04 instead.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -69,7 +69,7 @@ jobs:
 
   functionaltestsv1:
     name: Cgroup v1 Functional Tests
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


drakenclimber:
Thanks, @kamalesh-babulal.  Pushed to release-2.0 in commit 9460764acd4419885d126d3f8cc0a1fe5bc68159

PR #278, opened 2022-12-20
[RFC] cgroup v2 - Enable controllers at the leaf/child cgroup by default

kamalesh-babulal:
With cgroup v2, while creating a cgroup with a controller, the whole
hierarchy tree is walked, and the controller is enabled up to the parent
cgroup of the leaf node, this guarantees that the attached controller is
available in the leaf cgroups cgroup.controllers file, but this is
incomplete because the controller is not enabled in the leaf cgroup.
    
We do not have a separate API that is called to enable the controllers
but as a workaround, one can always call the `cgroup_set_values_recursive()`
to enable the controller in the leaf node.  To maintain the idea/compatibility
with the cgroup v1, where creating a cgroup on the controller mount point
or hierarchy is equivalent to enabling the controller in the cgroup v2, let's
enable the controller in the leaf node too.

This patch series can be divided into the api changes and gunit test changes.
Where the first two patches changes the api `_cgroup_create_cgroup()` and
`cgroupv2_controller_enabled()` respectively and patches 3,4 modifies the
unit testcases to match the changes to the APIs.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -2706,6 +2706,12 @@ static int _cgroup_create_cgroup(const struct cgroup * const cgroup,
 		goto err;
 
 	if (controller) {
+		if (version == CGROUP_V2) {
+			error = cgroupv2_subtree_control(base, controller->name, true);
+			if (error)
+				goto err;
+		}
+
 		error = cgroup_set_values_recursive(base, controller, false);
 		if (error)
 			goto err;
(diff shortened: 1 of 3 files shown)


drakenclimber:
This patchset has been merged in via pull request #281 and thus it can be closed.  Thanks, all.

PR #279, opened 2023-01-03
tools/cgx{get,set} fix tool name usage

kamalesh-babulal:
Both `cgxget` and `cgget` were forked from `cgget` and `cgset` while
adding the abstraction layer support.  They still retain the original tool's
name, which might get confusing to the users of the errors/warnings
displayed.  This patch replaces the warnings/errors/info displayed to
the user, along with non-history occurrences of the original tool names
with the newer tool name.

### DIFF
--- a/src/tools/cgxget.c
+++ b/src/tools/cgxget.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: LGPL-2.1-only
 /**
- * Libcgroup extended cgget.  Works with both cgroup v1 and v2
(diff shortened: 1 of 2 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commits 15fcda71fd6d and ddfc298084c2.  Pushed to release-3.0 in commits 7cc16ee6bbae and 588dd5e17dfa.

Thanks, @kamalesh-babulal

PR #280, opened 2023-01-04
tools/cgxset: fix help/usage formatting

kamalesh-babulal:
The help/usage output is misaligned, which is a mix of tabs and spaces.
Fix it by using only spaces for the alignment, so that we have a visual
representation of output in the code too, instead of having unaligned
info()'s needing to calculate the tabs stops always.

```
Before:
-------
Usage: cgxset [-r <name=value>] <cgroup_path> ...
   or: cgxset --copy-from <source_cgroup_path> <cgroup_path> ...
Set the parameters of given cgroup(s)
  -1, --v1                      Provided parameters are in v1 format
  -2, --v2                      Provided parameters are in v2 format
  -i, --ignore-unmappable       Do not return an error for settings that cannot be converted
  -r, --variable <name>                 Define parameter to set
  --copy-from <source_cgroup_path>      Control group whose parameters will be copied
```
```
After:
------
Usage: cgxset [-r <name=value>] <cgroup_path> ...
   or: cgxset --copy-from <source_cgroup_path> <cgroup_path> ...
Set the parameters of given cgroup(s)
  -1, --v1                                Provided parameters are in v1 format
  -2, --v2                                Provided parameters are in v2 format
  -i, --ignore-unmappable                 Do not return an error for settings that cannot be converted
  -r, --variable <name>                   Define parameter to set
  --copy-from <source_cgroup_path>        Control group whose parameters will be copied
```

### DIFF
--- a/src/tools/cgxset.c
+++ b/src/tools/cgxset.c
@@ -75,12 +75,12 @@ static void usage(int status, const char *program_name)
 	info("Usage: %s [-r <name=value>] <cgroup_path> ...\n", program_name);
 	info("   or: %s --copy-from <source_cgroup_path> <cgroup_path> ...\n", program_name);
 	info("Set the parameters of given cgroup(s)\n");
-	info("  -1, --v1                      Provided parameters are in v1 format\n");
-	info("  -2, --v2                      Provided parameters are in v2 format\n");
-	info("  -i, --ignore-unmappable       Do not return an error for ");
+	info("  -1, --v1                                Provided parameters are in v1 format\n");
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 44a90415da4ee96bee2d7b51673cd53fe54824a1.  Pushed to release-3.0 in commit 9f1ecd398a6715ba618d345eecf08c70101ee843.  Thanks, @kamalesh-babulal

Issue #272, opened 2022-12-01
cgroup_get_cgroup() broken for cgroupv2?


adriaan42:
I'm running libcgroup v3.0.0 on Linux 6.1:
```sh
# uname -a
Linux box 6.1.0-0-rt-amd64 #1 SMP PREEMPT_RT Debian 6.1~rc3-1~exp1 (2022-11-02) x86_64 GNU/Linux
# mount | grep cgroup
cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime,nsdelegate,memory_recursiveprot)
```

With the following reproducer:
```c
#include <stdlib.h>
#include <unistd.h>
#include <libcgroup.h>
#include "libcgroup-internal.h"

struct cgroup* get_or_create_cgroup(char *name)
{
    struct cgroup *cg = cgroup_new_cgroup(name);
    int ret = cgroup_get_cgroup(cg);
    if (ret == ECGROUPNOTEXIST) {
        struct cgroup_controller *cgc = cgroup_add_controller(cg, "cpuset");
        int ret2 = cgroup_create_cgroup(cg, 1);
        if (ret2) {
            printf("failed to create %s: %s\n", name, cgroup_strerror(ret2));
            exit(1);
        }
        printf("created %s\n", name);
    } else if (ret) {
        printf("failed to get %s: %s\n", name, cgroup_strerror(ret));
        exit(1);
    } else {
        printf("found existing %s\n", name);
#ifdef WORKAROUND
        cgroup_free(&cg);
        cg = cgroup_new_cgroup(name);
#endif
    }
    printf("  controllers:");
    for (int i = 0; i < CG_CONTROLLER_MAX; i++) {
        if (!cg->controller[i])
            break;
        printf(" %s", cg->controller[i]->name);
    }
    printf("\n");
    return cg;
}

int main(int, char*[]) {
    int ret = cgroup_init();
    if (ret) {
        printf("failed to initialize: %s\n", cgroup_strerror(ret));
        exit(1);
    }

    get_or_create_cgroup("my.slice");
    struct cgroup *cg = get_or_create_cgroup("my.slice/sub.slice");
    ret = cgroup_attach_task(cg);
    if (ret) {
        printf("failed to attach thread: %s\n", cgroup_strerror(ret));
        exit(1);
    }
    printf("success!\n");
}
```

And I'm seeing:
```sh
# gcc -o cgtest cgtest.c -lcgroup
[...]

PR #281, opened 2023-01-04
[BUG] Fix cgroup_get_cgroup() on cgroup v2 systems

drakenclimber:
Issue #272 rightfully raised a use case where `cgroup_get_cgroup()` doesn't work properly on cgroup v2.  This patchset fixes `cgroup_get_cgroup()` on v2 systems and adds a few tests around this functionality.

Note that one key change is that the requested controller(s) is(are) now enabled within the new cgroup.  Prior to this patchset the requested controller(s) was(were) only enabled in the parent cgroup's `cgroup.subtree_control` file.  The previous behavior was inconsistent with libcgroup cgroup v1 behavior.

(This needs to be evaluated for backport to the release-2.0 branch.)

### DIFF
--- a/include/libcgroup/groups.h
+++ b/include/libcgroup/groups.h
@@ -651,6 +651,31 @@ int cgroup_get_controller_version(const char * const controller,
 enum cg_setup_mode_t cgroup_setup_mode(void);
 
 /**
+ * Return the number of controllers for the specified cgroup in libcgroup
+ * internal structures.
+ *
+ * @param cgroup
+ * @return Count of the controllers or -1 on error.
+ */
+int cgroup_get_controller_count(struct cgroup *cgroup);
+
+/**
+ * Return requested controller from given group
+ *
+ * @param cgroup
+ * @param index The index into the cgroup controller list
+ */
+struct cgroup_controller *cgroup_get_controller_by_index(struct cgroup *cgroup, int index);
+
+/**
+ * Given a controller pointer, get the name of the controller
+ *
+ * @param controller
+ * @return controller name string, NULL if there's an error
+ */
+char *cgroup_get_controller_name(struct cgroup_controller *controller);
+
+/**
  * @}
  * @}
  */
--- a/src/api.c
+++ b/src/api.c
@@ -2706,6 +2706,12 @@ static int _cgroup_create_cgroup(const struct cgroup * const cgroup,
 		goto err;
 
 	if (controller) {
+		if (version == CGROUP_V2) {
+			error = cgroupv2_subtree_control(base, controller->name, true);
+			if (error)
+				goto err;
+		}
+
 		error = cgroup_set_values_recursive(base, controller, false);
 		if (error)
 			goto err;
--- a/src/api.c
+++ b/src/api.c
@@ -2277,17 +2277,22 @@ STATIC int cgroupv2_get_subtree_control(const char *path, const char *ctrl_name,
 	*enabled = false;
 
 	path_copy = (char *)malloc(FILENAME_MAX);
-	if (!path_copy)
+	if (!path_copy) {
+		error = ECGOTHER;
 		goto out;
+	}
 
 	ret = snprintf(path_copy, FILENAME_MAX, "%s/%s", path, CGV2_SUBTREE_CTRL_FILE);
-	if (ret < 0)
+	if (ret < 0) {
+		error = ECGOTHER;
 		goto out;
+	}
 
 	fp = fopen(path_copy, "re");
 	if (!fp) {
 		cgroup_warn("fopen failed\n");
 		last_errno = errno;
+		error = ECGOTHER;
 		goto out;
 	}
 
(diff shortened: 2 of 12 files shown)


drakenclimber:
Fixes Issue #272

src/api.c
```diff
@@ -3506,6 +3506,16 @@ int cgroup_get_cgroup(struct cgroup *cgroup)
 			cgroup->tasks_gid = stat_buffer.st_gid;
 
 			free(control_path);
+		} else { /* cgroup v2 */
+			bool enabled;
+
+			ret = cgroupv2_get_subtree_control(path, cg_mount_table[i].name,
+							  &enabled);
+			if (ret)
```
kamalesh-babulal, line 3514: @drakenclimber Thank you much for adding the Python bindings/test cases and the patchset looks great.  I have some minor comments.
Should we bail out on the error instead of continuing because the error might be due to `malloc()` or the `cgroup.subtree_control` not readable, given that `CGROUP V2` should always succeed in reading.

include/libcgroup/groups.h
```diff
@@ -650,6 +650,31 @@ int cgroup_get_controller_version(const char * const controller,
  */
 enum cg_setup_mode_t cgroup_setup_mode(void);
 
+/**
+ * Return the number of controllers for the specified cgroup in @c libcgroup
```
kamalesh-babulal, line 654: I did not understand `in @c ` part, sorry if I am missing the obvious.

tests/ftests/051-sudo-cgroup_get_cgroup.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: LGPL-2.1-only
+#
+# Exercise cgroup_create_cgroup() and cgroup_get_cgroup()
+#
+# Copyright (c) 2022 Oracle and/or its affiliates.
```
kamalesh-babulal, line 6: I  understand some of the test cases were written in  2022, should the year get updated to 2023?

tests/ftests/051-sudo-cgroup_get_cgroup.py
```diff
@@ -0,0 +1,95 @@
...
+
+    cg = Cgroup(CGNAME, Version.CGROUP_V2)
+    cg.get()
+
+    if len(cg.controllers) != 1:
+        # only one controller, cpu, should be enabled
+        result = consts.TEST_FAILED
+        cause = 'Expected one controller to be enabled, but {} were enabled'.format(
+                len(cg.controllers))
+        return result, cause
```
kamalesh-babulal, line 59: this return is followed by another return, though the scopes are different, this one can be removed.

tests/ftests/052-sudo-cgroup_attach_task.py
```diff
@@ -0,0 +1,102 @@
...
+        return result, cause
+
+    result = consts.TEST_PASSED
+    cause = None
+
+    return result, cause
+
+
+def setup(config):
+    print('setup')
```
kamalesh-babulal, line 43: Probably a statement from the debugging :)

tests/ftests/053-sudo-cgroup_attach_task_pid.py
```diff
@@ -0,0 +1,104 @@
...
+        return result, cause
+
+    result = consts.TEST_PASSED
+    cause = None
+
+    return result, cause
+
+
+def setup(config):
+    print('setup')
```
kamalesh-babulal, line 43: Probably a statement from the debugging :)

src/api.c
drakenclimber: Yeah, I think that's a good idea.  I think I had a bug or something in an early iteration of this patch, and I put the `continue` in to alleviate it.  It's no longer needed.

include/libcgroup/groups.h
drakenclimber: No clue what it is either; looks like my keyboard had a bad day :)

tests/ftests/051-sudo-cgroup_get_cgroup.py
drakenclimber: Yep

tests/ftests/051-sudo-cgroup_get_cgroup.py
drakenclimber: Sure

src/api.c
drakenclimber: Actually it was correct.  Controllers can be mounted v2 but not enabled in the subtree file, and thus they'll return `ECGROUPNOTMOUNTED`.

I think this is the right answer (along with some better return codes in `cgroupv2_get_subtree_control()`):
```
+               } else { /* cgroup v2 */
+                       bool enabled;
+
+                       error = cgroupv2_get_subtree_control(path, cg_mount_table[i].name,
+                                                            &enabled);
+                       if (error == ECGROUPNOTMOUNTED)
+                               continue;
+                       if (error)
+                               goto unlock_error;
+
+                       if (!enabled)
+                               continue;
```

drakenclimber: `cgroupv2_get_subtree_control()` is probably a good candidate for refactoring, though.  It only returns `enabled=1` when the error code is 0.  This is kinda silly and obviously redundant.

kamalesh-babulal: Agreed, as we spoke, let me take up the error code auditing.

kamalesh-babulal (approved):

drakenclimber:
Pushed to main and release-3.0.  Thanks, @kamalesh-babulal.

I'll start working on backporting this to release-2.0

PR #284, opened 2023-01-11
ftests/cgroup.py: add support to list shared mount point [v3.0+]

kamalesh-babulal:
`Cgroup::get_cgroup_mounts()`, could create a list of mount points and in
the case of shared mount points such as `cpu,cpuacct` in the cgroup v1. It
would only add a mount point for the last of the two controllers listed as
the mount point, for example:
`cgroup /sys/fs/cgroup/cpu,cpuacct cgroup rw,nosuid,nodev,noexec,relatime,cpu,cpuacct 0 0`

mount list will look over cpu controller.  Add support to recognize and
add both controllers to the list of mount points.

### DIFF
--- a/tests/ftests/cgroup.py
+++ b/tests/ftests/cgroup.py
@@ -20,7 +20,7 @@ import os
 
 
 class CgroupMount(object):
-    def __init__(self, mount_line):
+    def __init__(self, mount_line, controller=None):
         entries = mount_line.split()
 
         if entries[2] == 'cgroup':
@@ -32,14 +32,17 @@ class CgroupMount(object):
 
         self.mount_point = entries[1]
 
-        self.controller = None
-        if self.version == CgroupVersion.CGROUP_V1:
-            self.controller = entries[3].split(',')[-1]
+        if controller:
+            self.controller = controller
+        else:
+            self.controller = None
+            if self.version == CgroupVersion.CGROUP_V1:
+                self.controller = entries[3].split(',')[-1]
 
-            if self.controller == 'clone_children':
-                # the cpuset controller may append this option to the end
-                # rather than the controller name like all other controllers
-                self.controller = 'cpuset'
+                if self.controller == 'clone_children':
+                    # the cpuset controller may append this option to the end
+                    # rather than the controller name like all other controllers
+                    self.controller = 'cpuset'
 
     def __str__(self):
         out_str = 'CgroupMount'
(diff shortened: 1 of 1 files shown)


drakenclimber:
Good find.  Is this a simpler solution?  It removes the funky string stuff you had to do as well as removes the extra loop.
```
class CgroupMount(object):
-    def __init__(self, mount_line):
+    def __init__(self, mount_line, controller=None):
         entries = mount_line.split()
 
         if entries[2] == 'cgroup':
@@ -32,15 +32,18 @@ class CgroupMount(object):
 
         self.mount_point = entries[1]
 
-        self.controller = None
-        if self.version == CgroupVersion.CGROUP_V1:
-            self.controller = entries[3].split(',')[-1]
-
-            if self.controller == 'clone_children':
-                # the cpuset controller may append this option to the end
-                # rather than the controller name like all other controllers
-                self.controller = 'cpuset'
-
+        if controller:
+            self.controller = controller
+        else:
+            self.controller = None
+            if self.version == CgroupVersion.CGROUP_V1:
+                self.controller = entries[3].split(',')[-1]
+    
+                if self.controller == 'clone_children':
+                    # the cpuset controller may append this option to the end
+                    # rather than the controller name like all other controllers
+                    self.controller = 'cpuset'
+    

@@ -835,6 +838,13 @@ class Cgroup(object):
                 if mount.version == CgroupVersion.CGROUP_V1 or \
                    expand_v2_mounts is False:
                     mount_list.append(mount)
+
+                    if entry[1].find(',') > 0:
+                        # multiple controllers are mounted together.  Also add the
+                        # first controller to the mount_list
+                        controller = os.path.basename(entry[1].split(',')[0])
+                        mount = CgroupMount(line, controller=controller)
+                        mount_list.append(mount)
                     continue
```

kamalesh-babulal:
@drakenclimber Thank you for the review and the neat way to write it. I will re-spin the v2 with you as the author.

kamalesh-babulal:
**V2 Changes:**
- Adopted the re-write suggested by @drakenclimber (Tom Hromatka)

drakenclimber (approved):

drakenclimber:
Thanks, @kamalesh-babulal.  Pushed to main in commt ae6b2682f10e.  Pushed to release-3.0 in commit a881e7ab4eae.  This has not been pushed to release-2.0.

PR #285, opened 2023-01-23
Fix bug in cgroup_delete_cgroup_ext() [v3.0+]

kamalesh-babulal:
In cgroup v1, the controllers can share mount points, and when the
cgroup is deleted on one of the controller mount points (symlink), as
the side effect is deleted for the controllers sharing the mount point.
But this might not be true always, the user might try to `call cgroup_delete_cgroup_ext()`
with  a cgroup that doesn't exist over the share mount point and still
gets the return code to be zero.
    
This patch series attempts to fix this issue by introducing `cgrp_del_shared_mnt`
flag, which gets set when deleting the cgroup on the first of the controllers
sharing the mount point and checking the flag on the second controller mount
point to ensure that cgroup existed.

It also fixes the logic the in the testcase `ftests/047` to adopt the change in the
`cgroup_delete_cgroup_ext()` and adds a new testcases, that attempts to
delete non-existent cgroup on the shared mount point

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -2886,6 +2886,7 @@ static int cgroup_find_parent(struct cgroup *cgroup, char *controller, char **pa
 
 	if (stat(child_path, &stat_child) < 0) {
 		if (is_cgrp_ctrl_shared_mnt(controller)) {
+			last_errno = errno;
 			ret = ECGROUPNOTEXIST;
 			goto free_parent;
 		}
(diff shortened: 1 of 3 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commits 07259f1f17a3, 1179fa92a66e, and 01fc4107de92.  Pushed to release-3.0 in commits 5890504f47db, a33aaacccd40, and 7b205f245d03.  Thanks, @kamalesh-babulal

PR #286, opened 2023-01-25
fix segfault in cgroup_set_permissions() 

kamalesh-babulal:
This patch set fixes a segfault in the cgroup_set_permissions() and
also adds a simple fuzzer, that exercises the unlikely code path of
APIs by passing negative/wrong values to the APIs function
arguments. To begin with the fuzzer exercises the
`cgroup_set_permissions()` and gradually will add more APIs.


drakenclimber (approved):

drakenclimber:
Pushed to main in commits 9795cdf9c188 and dac22679ee3e.  Pushed to release-3.0 in commits 34e98305dd25 and 2611ac7f94c5.  Thanks, @kamalesh-babulal

Issue #297, opened 2023-02-13
RFE: Reading existing values of "cgroup.*" 


justinfx:
Hi there. I've been a user of libcgroup with cgroup v1 since 2013, and I am just now updating my custom python bindings to support the new v2 support. But I am struggling to understand the intended way to access "cgroup.*" values that are not specific to the traditional controllers.

This question is very much related to #131 where a similar question was asked regarding setting "cgroup.type". I had worked out from reading the source code that it is valid to add a "cgroup" controller and then I was able to set the value of "cgroup.freezer" to 1. However I am failing to understand how to initially read the values. It seems the "add_value" api wants to modify the value in addition to loading it into the internal structure. And it doesnt seem one can read a value unless its listed in the internal structure for the controller value names. But when I add the "cgroup" controller, the value count is 0.

So then, how does one read a value like "cgroup.freezer" or other "cgroup.*" values before they are written to?

drakenclimber:
Thanks for the question, @justinfx.  We have slowly added support for reading/writing `cgroup.*` files, but as you've noted, it's still lacking in some places.

For example, we really need to support a program like this:
```
#!/usr/bin/env python3
from libcgroup import Cgroup, Version

cg = Cgroup('user.slice', Version.CGROUP_V2)
cg.add_controller('cgroup')
cg.get()

print(cg)
```

But it's not there yet.  The above code will return all "real" controllers and their settings, but the `cgroup.*` files are not returned.  (Note that the exact syntax may change slightly, as I'm not sure if creating a controller called `cgroup` is the best way to go about things.)

I'll add this to our todo list for the v3.0.1 milestone.  Looks like `cgroup_get_cgroup()` needs to be made smarter to handle a cgroup named `cgroup`.  (That's a lot of cgroups :) .)  And as always, patches are more than welcome :)

drakenclimber:
Resolved in pull request #300

PR #300, opened 2023-02-15
Add support for reading cgroup.* files

drakenclimber:
Add support to `cgroup_get_cgroup()` to read disabled controllers and psuedo-controllers like cgroup.*.  Note that this change was only added to cgroup v2 and cgroup v1 remains the same.

Example usage:
```
#!/usr/bin/env python3
from libcgroup import Cgroup, Version

# Read all _enabled_ controllers.  The "cgroup" controller will not be read in this case
cg = Cgroup('user.slice', Version.CGROUP_V2)
cg.get()
print(cg)

# Read the cpuset controller.  This will work whether it's enabled or disabled
cg = Cgroup('user.slice', Version.CGROUP_V2)
cg.add_controller('cpuset')
cg.get()
print(cg)

# Read the "cgroup" controller
cg = Cgroup('user.slice', Version.CGROUP_V2)
cg.add_controller('cgroup')
cg.get()
print(cg)

# And of course, it's legal to combine controllers as you see fit
cg = Cgroup('user.slice', Version.CGROUP_V2)
cg.add_controller('cpuset')
cg.add_controller('cgroup')
cg.add_controller('memory')
cg.get()
print(cg)
```

Closes: #297

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -3554,6 +3554,7 @@ int cgroup_get_cgroup(struct cgroup *cgroup)
 	int initial_controller_cnt;
 	char *control_path = NULL;
 	char path[FILENAME_MAX];
+	int ctrl_cnt = 0;
 	DIR *dir = NULL;
 	int error;
 	int i, j;
@@ -3644,8 +3645,10 @@ int cgroup_get_cgroup(struct cgroup *cgroup)
 				 * and we've made it this far, then they are explicitly
 				 * interested in this controller and we should not remove it.
 				 */
-				if (initial_controller_cnt == 0)
+				if (initial_controller_cnt == 0) {
+					ctrl_cnt++;
 					continue;
+				}
 			} else if (error) {
 				goto unlock_error;
 			}
@@ -3666,6 +3669,8 @@ int cgroup_get_cgroup(struct cgroup *cgroup)
 			error = ECGOTHER;
 			goto unlock_error;
 		}
+
+		ctrl_cnt++;
 
 		while ((ctrl_dir = readdir(dir)) != NULL) {
 			/* Skip over non regular files */
@@ -3707,8 +3712,14 @@ int cgroup_get_cgroup(struct cgroup *cgroup)
 		}
 	}
 
-	/* Check if the group really exists or not */
-	if (!cgroup->index) {
+	/*
+	 * Check if the group really exists or not.  The cgroup->index controller count can't
+	 * be used in this case because cgroup v2 allows controllers to be enabled/disabled in
+	 * the subtree_control file.  Rather, cgroup_get_cgroup() tracks the number of possible
+	 * controllers in the ctrl_cnt variable and uses that to determine if the cgroup exists
+	 * or not.
+	 */
+	if (!ctrl_cnt) {
 		error = ECGROUPNOTEXIST;
 		goto unlock_error;
 	}
--- a/src/api.c
+++ b/src/api.c
@@ -3535,6 +3535,7 @@ fill_error:
 int cgroup_get_cgroup(struct cgroup *cgroup)
 {
 	struct dirent *ctrl_dir = NULL;
+	int initial_controller_cnt;
 	char *control_path = NULL;
 	char path[FILENAME_MAX];
 	DIR *dir = NULL;
@@ -3552,11 +3553,30 @@ int cgroup_get_cgroup(struct cgroup *cgroup)
 		return ECGROUPNOTALLOWED;
 	}
 
+	initial_controller_cnt = cgroup->index;
+
 	pthread_rwlock_rdlock(&cg_mount_table_lock);
 	for (i = 0; i < CG_CONTROLLER_MAX && cg_mount_table[i].name[0] != '\0'; i++) {
 		struct cgroup_controller *cgc;
 		struct stat stat_buffer;
 		int path_len;
+
+		if (initial_controller_cnt > 0) {
+			bool skip_this_controller = true;
+
+			/*
+			 * The user has specified a list of controllers they are interested
+			 * in.  Only operate on the specified controllers
+			 */
+			for (j = 0; j < cgroup->index; j++) {
+				if (strncmp(cg_mount_table[i].name, cgroup->controller[j]->name,
+					    CONTROL_NAMELEN_MAX) == 0)
+					skip_this_controller = false;
+			}
+
+			if (skip_this_controller)
+				continue;
+		}
 
 		if (!cg_build_path_locked(NULL, path, cg_mount_table[i].name))
 			continue;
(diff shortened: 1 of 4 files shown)


src/api.c
```diff
@@ -5267,6 +5283,18 @@ int cgroup_get_controller_next(void **handle, struct cgroup_mount_point *info)
 		goto out_unlock;
 	}
 
+	if (strncmp(cg_mount_table[*pos].name, CGROUP_FILE_PREFIX, CONTROL_NAMELEN_MAX) == 0)
```
drakenclimber, line 5286: This change was needed to appease `lscgroup`.  The `cgroup` controller was showing up in its lists of controllers, and I don't think this is desired behavior

kamalesh-babulal (approved):

drakenclimber:
@justinfx - Do you want to weigh in on these changes before they're merged?

justinfx:
@drakenclimber thanks for working on this!
It looks great to me from a code standpoint and I understand and agree with hiding the pseudo cgroup controller from listing.

I was wondering if this change might fix another problem I was seeing as well. I was finding that libcgroup seemed to report a group as not existing if I try to load one that has no subtree_control entries. As a specific example, if I created a parent group and configured controllers and then added a child "leaf" with no controllers so that I can add pids without the "no inner process" failure, libcgroup seemed to say "leaf" doesn't exist. I then had to at least add the "cpu" controller to the "leaf" subtree_control to make the group loadable. Does that sound like a problem that might exist and might be fixed by the changes here where it will load all controllers anyways?

drakenclimber:
> @drakenclimber thanks for working on this! It looks great to me from a code standpoint and I understand and agree with hiding the pseudo cgroup controller from listing.

Thanks!  Glad to help :)

> I was wondering if this change might fix another problem I was seeing as well. I was finding that libcgroup seemed to report a group as not existing if I try to load one that has no subtree_control entries. As a specific example, if I created a parent group and configured controllers and then added a child "leaf" with no controllers so that I can add pids without the "no inner process" failure, libcgroup seemed to say "leaf" doesn't exist. I then had to at least add the "cpu" controller to the "leaf" subtree_control to make the group loadable. Does that sound like a problem that might exist and might be fixed by the changes here where it will load all controllers anyways?

Do you have an example program that exhibits the problem?  I just ran the following on the HEAD of libcgroup (without the changes in this patchset), and it worked for me, but I'm guessing I am missing something.

```
#!/usr/bin/env python3
from libcgroup import Cgroup, Version
import time

parentcg = Cgroup('parent', Version.CGROUP_V2)
parentcg.create()

childcg = Cgroup('parent/child', Version.CGROUP_V2)
childcg.create()

childcg.attach()

time.sleep(100)
```

The sleep was simply there to give me time to explore around in bash and verify that the python process was present in `parent/child/cgroup.procs`

drakenclimber:
Oh, and do you want credit for reporting this issue, @justinfx?  If so, please provide the following:

`Reported-by: FirstName LastName <youremailaddress@somewhere.com>`

Thanks!

justinfx:
> Oh, and do you want credit for reporting this issue, @justinfx? If so, please provide the following:

Sure!
`Reported-by: Justin Israel <justinisrael@gmail.com>`

> Do you have an example program that exhibits the problem?

Actually, yes! Your sample script can reproduce it with a small tweak to make it match my actual use case:

```
#!/usr/bin/env python3
from libcgroup import Cgroup, Version

parentcg = Cgroup('parent', Version.CGROUP_V2)
parentcg.create()

childcg = Cgroup('parent/child', Version.CGROUP_V2)
childcg.create()

childcg2 = Cgroup('parent/child', Version.CGROUP_V2)
childcg2.get()
# RuntimeError: cgroup_get_cgroup failed: 50002
```

The problem occurs when you have an existing child cgroup that has no controllers enabled because the parent did not put anything into its own `cgroup.subtree_control`. But the following does work:

```
#!/usr/bin/env python3
from libcgroup import Cgroup, Version

parentcg = Cgroup('parent', Version.CGROUP_V2)
parentcg.add_controller('cpu')
parentcg.add_controller('memory')
parentcg.add_controller('cpuset')
parentcg.create()

childcg = Cgroup('parent/child', Version.CGROUP_V2)
# need to give child at least one controller allowed from parent,
# in its own subtree_control
childcg.add_controller('cpu')
childcg.create()

childcg2 = Cgroup('parent/child', Version.CGROUP_V2)
childcg2.get()
```

justinfx:
@drakenclimber I've tested this branch and it seems my other issue is actually not fixed by this patch. Although I am not clear if it should be, or if its entirely unrelated and should be reported in a separate ticket? The reason it felt related is because it felt like your changes were indeed trying to address the issue of not loading all controllers, and this second issue being related to not having any controllers enabled and resulting in the cgroup acting like it doesn't exist.

drakenclimber:
> @drakenclimber I've tested this branch and it seems my other issue is actually not fixed by this patch. Although I am not clear if it should be, or if its entirely unrelated and should be reported in a separate ticket? The reason it felt related is because it felt like your changes were indeed trying to address the issue of not loading all controllers, and this second issue being related to not having any controllers enabled and resulting in the cgroup acting like it doesn't exist.

Thanks for the feedback.  This is a tough balancing act we're trying to walk, and I'm not exactly sure what the right answer is.  We're colliding with a few different things - C expects one thing, Python another, and of course cgroup v1 and v2 behave very differently.

Recently, a similar question [1] regarding the C version of `cgroup_get_cgroup()` came in.  We changed how we were writing to the subtree_control file and we updated `cgroup_get_cgroup()` to report only the enabled controllers.

Hmmm... what would you expect libcgroup to return in a case like this:
```
sudo mkdir -p /sys/fs/cgroup/foo/bar
```
```
#!/usr/bin/env python3
from libcgroup import Cgroup, Version

cg = Cgroup('foo/bar', Version.CGROUP_V2)
cg.get()
print(cg)
```

I don't think RuntimeError is correct, but it seems reasonable that nothing is returned from `cg.get()`.  Thoughts?

[1] https://github.com/libcgroup/libcgroup/issues/272

drakenclimber:
Alright, I thought about it for a bit and added one small patch.  Let me know what you think, @justinfx.  It makes the above scenario work for me.

Here's the new patch:
https://github.com/drakenclimber/libcgroup/commit/604a74171ea53d706aca2a133a3c5ae94982116b

justinfx:
This patch works great! And I think it's the right thing to do. I can now get back an actual object for the group, and see that it has no enabled controllers. But I can then have access to stats, properties, proc lists, etc. It makes sense to reflect the group as existing, because pids are allowed to live in that child cgroup regardless of not having any controllers enabled. Really, those controllers in subtree_control are for subtree delegation. A leaf group may not have any enabled so that it can contain the pids for a parent group that does have controllers enabled.

drakenclimber:
> This patch works great! And I think it's the right thing to do. I can now get back an actual object for the group, and see that it has no enabled controllers. But I can then have access to stats, properties, proc lists, etc. It makes sense to reflect the group as existing, because pids are allowed to live in that child cgroup regardless of not having any controllers enabled. Really, those controllers in subtree_control are for subtree delegation. A leaf group may not have any enabled so that it can contain the pids for a parent group that does have controllers enabled.

Awesome.  And I completely agree with your analysis.

Thanks for testing it out!  I also added your `Reported-by:` to the latest version of the patchset

drakenclimber:
Pushed to the main and release-3.0 branches.  Thanks, @justinfx and @kamalesh-babulal.

PR #301, opened 2023-02-17
[RFC] tests/ftests: synchronize between github runners

kamalesh-babulal:
If a VM is shared between multiple github runners, there are chances of
each other stomping over other's run, if executed parallelly.  To avoid the
race between the runners, introduce lock file, that gets acquired (created)
when `ftest.sh` starts and gets removed by `ftest-nocontainer.sh`, this
ensures that both test cases are executed before other runners, that's
waiting for their chance to run.  A runner would wait for 10 minutes before
re-trying to run.  At the max, a runner would wait for 50 minutes (5 retries)
before giving up.

### DIFF
--- a/tests/ftests/ftests-nocontainer.sh
+++ b/tests/ftests/ftests-nocontainer.sh
@@ -6,6 +6,22 @@ AUTOMAKE_HARD_ERROR=99
 
 START_DIR=$PWD
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+RUNNER_LOCK_FILE="/var/run/libcgroup/github-runner.lock"
+
+# the lock file is removed after executing no-container
+# test cases, whereas the lock is taken by the ftests.sh
+function cleanup()
+{
+	sudo rm -f "$RUNNER_LOCK_FILE"
+	exit "$1"
+}
+
+# ftests.sh should have taken the lock, if not
+# something is out of order, let's abort
+if [ ! -f "$RUNNER_LOCK_FILE" ]; then
+	echo "Lock file $RUNNER_LOCK_FILE missing, aborting"
+	exit 1
+fi
 
 if [ "$START_DIR" != "$SCRIPT_DIR" ]; then
 	cp "$SCRIPT_DIR"/*.py "$START_DIR"
@@ -21,10 +37,10 @@ fi
 	-n Libcg"$RANDOM"
 RET1=$?
 
-pushd ../../src || exit $AUTOMAKE_HARD_ERROR
+pushd ../../src || cleanup $AUTOMAKE_HARD_ERROR
 PATH="$PATH:$(pwd)"
 export PATH
-popd || exit $AUTOMAKE_HARD_ERROR
+popd || cleanp $AUTOMAKE_HARD_ERROR
 
 sudo PATH=$PATH PYTHONPATH=$PYTHONPATH ./ftests.py -l 10 -s "sudo" \
 	-L "$START_DIR/ftests-nocontainer.py.sudo.log" --no-container -n Libcg"$RANDOM"
@@ -40,20 +56,20 @@ fi
 
 if [[ $RET1 -ne $AUTOMAKE_SKIPPED ]] && [[ $RET1 -ne 0 ]]; then
 	# always return errors from the first test run
-	exit $RET1
+	cleanup $RET1
 fi
 if [[ $RET2 -ne $AUTOMAKE_SKIPPED ]] && [[ $RET2 -ne 0 ]]; then
 	# return errors from the second test run
-	exit $RET2
+	cleanup $RET2
 fi
 
 if [[ $RET1 -eq 0 ]] || [[ $RET2 -eq 0 ]]; then
-	exit 0
+	cleanup 0
 fi
 
 if [[ $RET1 -eq $AUTOMAKE_SKIPPED ]] || [[ $RET2 -eq $AUTOMAKE_SKIPPED ]]; then
-	exit $AUTOMAKE_SKIPPED
+	cleanup $AUTOMAKE_SKIPPED
 fi
 
 # I don't think we should ever get here, but better safe than sorry
-exit $AUTOMAKE_HARD_ERROR
+cleanup $AUTOMAKE_HARD_ERROR
--- a/tests/ftests/ftests.sh
+++ b/tests/ftests/ftests.sh
@@ -4,6 +4,14 @@
 START_DIR=$PWD
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
+# synchronize between different github runners running on
+# same VM's, this will stop runners from stomping over
+# each other's run.
+LIBCGROUP_RUN_DIR="/var/run/libcgroup/"
+RUNNER_LOCK_FILE="/var/run/libcgroup/github-runner.lock"
+RUNNER_SLEEP_SECS=300		# sleep for 5 minutes
+RUNNER_MAX_TRIES=10		# Abort after 50 minutes, if we don't chance to run
+
 if [ "$START_DIR" != "$SCRIPT_DIR" ]; then
 	cp "$SCRIPT_DIR"/*.py "$START_DIR"
 fi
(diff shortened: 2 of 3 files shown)


tests/ftests/ftests-nocontainer.sh
```diff
@@ -21,10 +37,10 @@ fi
 	-n Libcg"$RANDOM"
 RET1=$?
 
-pushd ../../src || exit $AUTOMAKE_HARD_ERROR
+pushd ../../src || cleanup $AUTOMAKE_HARD_ERROR
 PATH="$PATH:$(pwd)"
 export PATH
-popd || exit $AUTOMAKE_HARD_ERROR
+popd || cleanp $AUTOMAKE_HARD_ERROR
```
drakenclimber, line 43: typo - should be cleanup :)

tests/ftests/ftests-nocontainer.sh
```diff
@@ -6,6 +6,22 @@ AUTOMAKE_HARD_ERROR=99
 
 START_DIR=$PWD
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+RUNNER_LOCK_FILE="/var/run/libcgroup/github-runner.lock"
```
drakenclimber, line 9: I can see the lock file being needed for `ftests-nocontainer.sh` since the tests are operating on the "bare-metal" `/sys/fs/cgroup`.  But is it needed for `ftests.sh`?  `ftests.sh` generates a random container name for each run, so I don't think container foo should collide with container bar.

https://github.com/libcgroup/libcgroup/blob/main/tests/ftests/ftests.sh#L17

drakenclimber: Or are we doing this because the VMs are underpowered and they can't support simultaneous test runs?

tests/ftests/ftests.sh
```diff
@@ -4,6 +4,14 @@
 START_DIR=$PWD
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
+# synchronize between different github runners running on
+# same VM's, this will stop runners from stomping over
+# each other's run.
+LIBCGROUP_RUN_DIR="/var/run/libcgroup/"
+RUNNER_LOCK_FILE="/var/run/libcgroup/github-runner.lock"
+RUNNER_SLEEP_SECS=600		# sleep for 10 minutes
```
drakenclimber, line 12: 10 minutes feels like a long wait time to me, but I haven't thought about it much :/

drakenclimber:
Looks like a cool idea to me.  I added a few questions, but nothing major.

It does feel weird that `ftests.sh` grabs the lock and `ftests-nocontainer.sh` releases the lock.  Seems like a weird coupling between the two.

tests/ftests/ftests-nocontainer.sh
kamalesh-babulal: @drakenclimber Thank you so much for the review.  Yes, it's because of the underpowered VMs.

tests/ftests/ftests.sh
kamalesh-babulal: An average run takes between 10 to 15 minutes to complete, hence the lower bound of 10 minutes.

tests/ftests/ftests-nocontainer.sh
kamalesh-babulal: Sorry, it should have been `cleaup`. Will fix it up in the next iteration.

kamalesh-babulal:
**V2 Changes**
- Reduce the timeout to 5 minutes and increases the max tries to 10 (suggested by @drakenclimber)
- Introduce timeout for races between `apt` operations.

drakenclimber:
I have a bad feeling that splitting the lock/unlock into separate files will bite us.  I've reworked the locking patch into this:
https://github.com/drakenclimber/libcgroup/commit/68a4fba70a658b036387bf4fb1664842126e4f93

What do you think, @kamalesh-babulal?

kamalesh-babulal:
> I have a bad feeling that splitting the lock/unlock into separate files will bite us. I've reworked the locking patch into this:
https://github.com/drakenclimber/libcgroup/commit/68a4fba70a658b036387bf4fb1664842126e4f93

@drakenclimber I liked your approach, it's cleaner.  `shellcheck`  reported a couple of warnings on the `ftest-wrapper.sh` (ran it locally) and I have fixed them, please feel free to include them:
```
diff --git a/tests/ftests/ftests-wrapper.sh b/tests/ftests/ftests-wrapper.sh
index 10e1c04..5a76c5d 100755
--- a/tests/ftests/ftests-wrapper.sh
+++ b/tests/ftests/ftests-wrapper.sh
@@ -26,10 +26,12 @@ if [ "$START_DIR" != "$SCRIPT_DIR" ]; then
        cp "$SCRIPT_DIR"/*.py "$START_DIR"
 fi
 
-if [ -d ../../src/python/build/lib.* ]; then
-       pushd ../../src/python/build/lib.*
-       export PYTHONPATH="$PYTHONPATH:$(pwd)"
-       popd
+PYTHON_LIBRARY_PATH="../../src/python/build/lib.*"
+if [ -d "$PYTHON_LIBRARY_PATH" ]; then
+       pushd "$PYTHON_LIBRARY_PATH" || cleanup $AUTOMAKE_HARD_ERROR
+       PYTHONPATH="$PYTHONPATH:$(pwd)"
+       export PYTHONPATH
+       popd || cleanup $AUTOMAKE_HARD_ERROR
 fi
 
 # If other runners are running then the file exists
@@ -66,7 +68,7 @@ PATH="$PATH:$(pwd)"
 export PATH
 popd || cleanup $AUTOMAKE_HARD_ERROR
 
-sudo PATH=$PATH PYTHONPATH=$PYTHONPATH ./ftests.py -l 10 -s "sudo" \
+sudo PATH="$PATH" PYTHONPATH="$PYTHONPATH" ./ftests.py -l 10 -s "sudo" \
        -L "$START_DIR/ftests-nocontainer.py.sudo.log" --no-container -n Libcg"$RANDOM"
 RET3=$?
 
```
in line 89:
```
   89 if [[ $RET1 -eq 0 ]] || [[ $RET2 -eq 0 ]] || [[ $RET3 -eq 0 ]]; then
   90         # at least one test passed and _no_ tests failed
   91         cleanup 0
   92 fi
```

consider this if any of the `RET?` is set to some non-zero value (failed test count), the test will
return a false positive.  A check to ensure that `RET` is `$AUTOMAKE_SKIPPED` when it's>0 is
required.  I am a little  biased by the way you handled it in `ftest-nocontainer.sh`

drakenclimber:
> @drakenclimber I liked your approach, it's cleaner. `shellcheck` reported a couple of warnings on the `ftest-wrapper.sh` (ran it locally) and I have fixed them, please feel free to include them:
> 

Thanks.  Will do.

> 
> in line 89:
> 
> ```
>    89 if [[ $RET1 -eq 0 ]] || [[ $RET2 -eq 0 ]] || [[ $RET3 -eq 0 ]]; then
>    90         # at least one test passed and _no_ tests failed
>    91         cleanup 0
>    92 fi
> ```
> 
> consider this if any of the `RET?` is set to some non-zero value (failed test count), the test will return a false positive. A check to ensure that `RET` is `$AUTOMAKE_SKIPPED` when it's>0 is required. I am a little biased by the way you handled it in `ftest-nocontainer.sh`

Oh, good catch.  I forgot that it returns the `failed_cnt`.  The logic in `ftest-nocontainer` worked well for two tests, but it was unwieldy for three.  Thus the change.

Let me think about it a bit.  I'll post an update today.

drakenclimber:
Alright, here's the update:
https://github.com/drakenclimber/libcgroup/commit/cffcbaad7cdca7353d00804d152bf82485d6548f

It was much easier to add the ftests-nocontainer `RET` logic than I thought.  Good call.

I did have to modify the PYTHON_LIBRARY_PATH logic a little bit.  Since we're wildcarding the path, we can't put it in quotes.

kamalesh-babulal:
> Alright, here's the update:
https://github.com/drakenclimber/libcgroup/commit/cffcbaad7cdca7353d00804d152bf82485d6548f
It was much easier to add the ftests-nocontainer RET logic than I thought. Good call.

@drakenclimber Looks good to me :)

> I did have to modify the PYTHON_LIBRARY_PATH logic a little bit. Since we're wildcarding the path, we can't put it in quotes.

`shellcheck` is unhappy with the way we are using glob:
```
# shellcheck ftests-wrapper.sh 

In ftests-wrapper.sh line 29:
PYTHON_LIBRARY_PATH=../../src/python/build/lib.*
                    ^--------------------------^ SC2125 (warning): Brace expansions and globs are literal in assignments. Quote it or use an array.


In ftests-wrapper.sh line 30:
if [ -d $PYTHON_LIBRARY_PATH ]; then
        ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
if [ -d "$PYTHON_LIBRARY_PATH" ]; then


In ftests-wrapper.sh line 31:
        pushd $PYTHON_LIBRARY_PATH || cleanup $AUTOMAKE_HARD_ERROR
              ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
        pushd "$PYTHON_LIBRARY_PATH" || cleanup $AUTOMAKE_HARD_ERROR

For more information:
  https://www.shellcheck.net/wiki/SC2125 -- Brace expansions and globs are li...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

kamalesh-babulal:
`shellcheck` seems to be okay with:
````
diff --git a/tests/ftests/ftests-wrapper.sh b/tests/ftests/ftests-wrapper.sh
index b7fd64b..95ec0cd 100755
--- a/tests/ftests/ftests-wrapper.sh
+++ b/tests/ftests/ftests-wrapper.sh
@@ -26,9 +26,9 @@ if [ "$START_DIR" != "$SCRIPT_DIR" ]; then
        cp "$SCRIPT_DIR"/*.py "$START_DIR"
 fi
 
-PYTHON_LIBRARY_PATH=../../src/python/build/lib.*
-if [ -d $PYTHON_LIBRARY_PATH ]; then
-       pushd $PYTHON_LIBRARY_PATH || cleanup $AUTOMAKE_HARD_ERROR
+PYTHON_LIBRARY_PATH=(../../src/python/build/lib*)
+if [ -d  "${PYTHON_LIBRARY_PATH[0]}" ]; then
+       pushd "${PYTHON_LIBRARY_PATH[0]}" || cleanup $AUTOMAKE_HARD_ERROR
        PYTHONPATH="$PYTHONPATH:$(pwd)"
        export PYTHONPATH
        popd || cleanup $AUTOMAKE_HARD_ERROR

drakenclimber:
Resolved in pull request #310.

PR #302, opened 2023-02-17
tools/cg{set,xset}: fix Uninitialized pointers read 

kamalesh-babulal:
This patch set fixes the fix Uninitialized pointers read, reported
by the Coverity tool in `tools/cgset` and `tools/cgxset`.

### DIFF
--- a/src/tools/cgset.c
+++ b/src/tools/cgset.c
@@ -131,8 +131,8 @@ int main(int argc, char *argv[])
 	int nv_max = 0;
 
 	char src_cg_path[FILENAME_MAX] = "\0";
-	struct cgroup *src_cgroup;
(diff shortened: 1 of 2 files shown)


drakenclimber:
Thanks, Kamalesh.  Good fixes.

I noticed that there are a couple ways that `converted_src_cgroup` could also not be freed.  I'll add them in a separate pull request.

drakenclimber (approved):

drakenclimber:
Pushed to main in commits 4611135a0672 and 94ba8b65f8b3.  Pushed to release-3.0 in commits 5cf24f84f31c and 9ba3b163a756.  Thanks, @kamalesh-babulal

PR #303, opened 2023-02-17
BUG: tools/cgxset: Fix potential memory leak in converted_src_cgroup

drakenclimber:
There's a brief window where converted_src_cgroup is allocated but hasn't been assigned to the *cgroup pointer.  If a failure occurs in this window, then converted_src_cgroup must be freed before goto'ing to the error path.

### DIFF
--- a/src/tools/cgxset.c
+++ b/src/tools/cgxset.c
@@ -272,7 +272,7 @@ int main(int argc, char *argv[])
 		if (!cgroup) {
 			ret = ECGFAIL;
 			err("%s: can't add new cgroup: %s\n", argv[0], cgroup_strerror(ret));
-			goto cgroup_free_err;
+			goto err;
 		}
 
 		/* copy the values from the source cgroup to new one */
@@ -280,7 +280,7 @@ int main(int argc, char *argv[])
 		if (ret != 0) {
 			err("%s: cgroup %s error: %s\n", argv[0], src_cg_path,
 			    cgroup_strerror(ret));
-			goto cgroup_free_err;
+			goto err;
 		}
 
 		converted_src_cgroup = cgroup_new_cgroup(cgroup->name);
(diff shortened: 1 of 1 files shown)


src/tools/cgxset.c
```diff
@@ -289,6 +289,11 @@ int main(int argc, char *argv[])
 			goto err;
```
kamalesh-babulal, line 289: @drakenclimber good catch.  One improvement would be to replace the goto `err`  with goto `cgroup_free_err`.   The `cgroup` is allocated and needs to free'd too.

src/tools/cgxset.c
```diff
@@ -298,8 +303,14 @@ int main(int argc, char *argv[])
 			 * v1 to v2 or vice versa
 			 */
 			ret = 0;
-		else if (ret)
+		else if (ret) {
+			free(converted_src_cgroup);
 			goto err;
```
kamalesh-babulal, line 308: This goto `err` too needs to be replaced with `cgroup_free_err`

drakenclimber: Good catch.  Looking through the function, I don't see the point of the `err` tag.  I propose we delete the `err` tag and have all the goto's go to `cgroup_free_err`.  Thoughts?

drakenclimber: It was lazier (easier) to delete the `cgroup_free_tag` and have everything goto `err`.  Let me know what you think.

drakenclimber:
@kamalesh-babulal - Would you mind taking a quick look at my latest changes and make sure I didn't screw anything up?  ;)  Thanks!

src/tools/cgxset.c
kamalesh-babulal: > Looking through the function, I don't see the point of the err tag. I propose we delete the err tag and have all the goto's go to cgroup_free_err. Thoughts?

Agreed, there is no point in having an extra label for the error path, given that there are checks for their validity before calling `cgroup_free()` and at the maximum, extra if checks will be performed in the error path.

kamalesh-babulal:
> Would you mind taking a quick look at my latest changes and make sure I didn't screw anything up? ;) Thanks!

@drakenclimber the changes look good to me and apologies for missing the update.

kamalesh-babulal (approved):

drakenclimber:
Pushed to main in commit f1579cec2734.  Pushed to release-3.0 in commit 8da741e19bb5.  Thanks, @kamalesh-babulal

PR #304, opened 2023-02-21
wrapper: fix segfault in cgroup_{set,get}_value_string() 

kamalesh-babulal:
This patch series fixes a segfault in `cgroup_set_value_string()` and
`cgroup_get_value_string()` APIs.  when `NULL` is passed in place of
the controller setting name or value or both. It also adds test cases
to the gunit fuzzer.


drakenclimber (approved):

drakenclimber:

PR #305, opened 2023-02-22
api: fix cgroup_set_value_string to use CG_CONTROL_VALUE_MAX instead of CG_VALUE_MAX

justinfx:
I've been maintaining this patch in my own vendored copy of libcgroup for many many years, and decided to push it up as a PR.

The value of CG_VALUE_MAX=100 is fine for the common case. But there is a case when it will cause a problem. On machines containing a large number of cpu ids, it is possible to end up with cpuset.cpus configurations that exceed 100 characters. This leads to a truncation of the value and assigning either less than expected, or a syntax error, depending on the truncation.

An example would be a host containing 64 cpu ids. If one wanted to assign every other cpu id, they would generate a cpuset.cpus value size of 181 bytes. In my own experience I have worked with hosts with 128 and 256 cpus and hit this issue many times, where the value wants to be well over 512 bytes.

My local patch has always been to set the value of CG_VALUE_MAX to 1024. But I realise that it may not be desirable to default every single cgroup_stat.value to such a large initial size, so I am submitting the patch at 512.

Another option might be to make a more granular change to special casing certain values that are known to support much larger value sizes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+#
+# Dependabot Workflow for libcgroup
+#
+# Copyright (c) 2023 Oracle and/or its affiliates.
+# Author: Tom Hromatka <tom.hromatka@oracle.com>
+#
+
+# based on this guide from GitHub:
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
+    commit-message:
+      prefix: "RFE: "
+    labels:
+      - "enhancement"
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,28 @@
+# SPDX-License-Identifier: LGPL-2.1-only
+#
+# Continuous Integration Workflow for libcgroup
+#
+# Copyright (c) 2020-2023 Oracle and/or its affiliates.
+# Author: Tom Hromatka <tom.hromatka@oracle.com>
+# Author: Kamalesh Babulal <kamalesh.babulal@oracle.com>
+#
+
+name: "CodeQL"
+on: ["push", "pull_request"]
+
+jobs:
+  codeql:
+    name: CodeQL
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: false
+    - uses: github/codeql-action/init@v2
+      with:
+        languages: cpp, python
+    - name: Initialize the directory
+      uses: ./.github/actions/setup-libcgroup
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -10,22 +10,6 @@ name: Continuous Integration
 on: ["push", "pull_request"]
 
 jobs:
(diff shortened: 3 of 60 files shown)


kamalesh-babulal:
Hi Justin,

Thank you for the patch. Previously the  `struct cgroup_controller.values` size  was increased to [ 4096 bytes]( https://github.com/libcgroup/libcgroup/blob/main/src/libcgroup-internal.h#L86), from `CG_VALUE_MAX`.  Increasing `cgroup_stat.value` to `512`, would mean having two conflicting sizes for the controller values, which we already have but `cgroup_controller` is internal only.  As far I can guess,  `struct cgroup_controller` might allocated more than `struct cgroup_stat` by the users but internal it seems right:
```
# git grep 'struct cgroup_controller' -- '*.[ch]'|wc -l
129
# git grep 'struct cgroup_stat' -- '*.[ch]'|wc -l
7
```
I might be completly wrong on the user usage, because I have not statistical evidience to prove this.  I guess, it should be okay to increase the size of `cgroup_stat.value` to `4096` too.  Given that we already have for `struct cgroup_controller`. One approach in the future would be to avoid static allocation and dynamically allocate the `cgroup_controller` and `cgroup_stat` value based on `get_nprocs()` x 2.

justinfx:
@kamalesh-babulal thanks for the clarification! It is possible that the state of the code today has wildly changed from back when I first created this patch, and it may no longer have the same degree of impact. From what you have pointed out, I have looked at the source and see that `control_value.value` (now 4096) is used in the public `cgroup_get_value_*` public API. Whereas `cgroup_stat.value` (100 -> 512) is used in the `cgroup_read_stats_*` public API. 

That being said, I think I actually see where a more specific "bug" is, related to this patch:
https://github.com/libcgroup/libcgroup/blob/08bb9a32592079b982e93b1e21fc084431334c2d/src/wrapper.c#L472

`cgroup_set_value_string` is using `CG_VALUE_MAX` to copy the value into the `control_value.value`, where I think it should actually be using `CG_CONTROL_VALUE_MAX`. This bug would be the source of my original problem when setting a very large cpuset, as the value would be truncated smaller than what the value field really allows. Do you think we should adjust this patch only to fix `cgroup_set_value_string`? I think maybe changing the `CG_VALUE_MAX` was only to indirectly fix this one spot, as I don't think stats files (multi-line value files) have very long lines.

kamalesh-babulal:
>  It is possible that the state of the code today has wildly changed from back when I first created this patch, and it may no longer have the same degree of impact.

Can you please share/upstream the local patches you have been carrying, I will try and help in getting them merged.
>This bug would be the source of my original problem when setting a very large cpuset, as the value would be truncated smaller than what the value field really allows. Do you think we should adjust this patch only to fix cgroup_set_value_string?

` cgroup_set_value_string` was missed in the original patch. Please, do send the patch to fix it.  `cgroup_stats` can be revisited when it breaks.

justinfx:
> Can you please share/upstream the local patches you have been carrying, I will try and help in getting them merged.

My original patch was exactly what the first commit was in this merge. It was very simply increasing `CG_VALUE_MAX`. But what I meant about the state of the code changing was that at the time I applied it, the `CG_VALUE_MAX` maybe had a more direct impact on what I was trying to fix. Whereas now there have been many patches since, including `CG_CONTROL_VALUE_MAX` and internal structures. So it makes more sense to use a different and more targeting fix at this point.

>  cgroup_set_value_string was missed in the original patch. Please, do send the patch to fix it. cgroup_stats can be revisited when it breaks.

I've reverted my `CG_VALUE_MAX` patch as it is no longer the best fix. My new commit now just replaces `CG_VALUE_MAX` with `CG_CONTROL_VALUE_MAX`, only in the `cgroup_set_value_string` function. Thanks!

justinfx:
Replaced by #307 with updated branches

PR #306, opened 2023-02-23
tests/ftest: use string filter() to concatenation cause

kamalesh-babulal:
if, then, else construction is used to concatenate the 'cause' string, that 
appends the reason for the test failures.  The reason to use the checks
to `TypeError` that occurs while concatenating `None` and `str` type.
Replace them with a one-liner `string.filter()`, across the test sources.


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 5d4bd5c2532a.  Pushed to release-3.0 in commit 7edd97c30dbc.  Thanks, @kamalesh-babulal

PR #307, opened 2023-02-23
api: cgroup_set_value_string: allocate correct value size to copy value provided 

justinfx:
Replaces libcgroup/libcgroup#305

`cgroup_set_value_string` is using CG_VALUE_MAX to copy the value into the `control_value.value`, where I think it should actually be using `CG_CONTROL_VALUE_MAX`. 

On machines containing a large number of cpu ids, it is possible to end up with `cpuset.cpus` configurations that exceed 100 characters. This leads to a truncation of the value and assigning either less than expected, or a syntax error, depending on the truncation.

An example would be a host containing 64 cpu ids. If one wanted to assign every other cpu id, they would generate a `cpuset.cpus` value size of 181 bytes. In my own experience I have worked with hosts with 128 and 256 cpus and hit this issue many times, where the value wants to be well over 512 bytes.

### DIFF
--- a/src/wrapper.c
+++ b/src/wrapper.c
@@ -469,7 +469,7 @@ int cgroup_set_value_string(struct cgroup_controller *controller, const char *na
 		struct control_value *val = controller->values[i];
 
 		if (!strcmp(val->name, name)) {
-			strncpy(val->value, value, CG_VALUE_MAX);
+			strncpy(val->value, value, CG_CONTROL_VALUE_MAX);
 			val->value[sizeof(val->value)-1] = '\0';
 			val->dirty = true;
 			return 0;


drakenclimber:
Thanks again for your submission, @justinfx.  I'll let @kamalesh-babulal weigh in on the patch itself, but I do have one minor request before he comes back online.

Since we are an open-source project, we need every submitter to own their work.  Could you please sign off on your commit?  See these links for more details:
https://github.com/libcgroup/libcgroup/blob/main/CONTRIBUTING.md#sign-your-work
https://developercertificate.org/

justinfx:
Sorry about that @drakenclimber! I've force pushed the commit to my branch with the amended signature.

kamalesh-babulal (approved):

drakenclimber:
Pushed to  main in commit b90efb3e2d77.  Pushed to release-3.0 in commit 3efb9fc5bc3f.  Thanks @justinfx.

PR #308, opened 2023-02-23
ftests: Cleanup pid logic in functional tests

drakenclimber:
PIDs were inconsistently being managed in the functional tests.  Some functions treated them as `int` and in others they were `str`.  Modify all functions that deal with PIDs to treat them as `int` or a list of `int`s.

### DIFF
--- a/tests/ftests/049-sudo-systemd_create_scope.py
+++ b/tests/ftests/049-sudo-systemd_create_scope.py
@@ -9,9 +9,10 @@
 
 from cgroup import Cgroup as CgroupCli
 from cgroup import CgroupVersion
-from run import Run, RunError
 from libcgroup import Cgroup
 from systemd import Systemd
+from process import Process
+from run import RunError
 import ftests
 import consts
 import sys
@@ -62,8 +63,7 @@ def test(config):
 def teardown(config, result):
     pid = CgroupCli.get(config, cgname=os.path.join(SLICE, SCOPE), setting='cgroup.procs',
                         print_headers=False, values_only=True)
-    if pid is not None:
-        Run.run(['kill', '-9', str(pid)], shell_bool=True)
+    Process.kill(config, pid)
 
     if result != consts.TEST_PASSED:
         # Something went wrong.  Let's force the removal of the cgroups just to be safe.
--- a/tests/ftests/065-sudo-systemd_cgclassify-v1.py
+++ b/tests/ftests/065-sudo-systemd_cgclassify-v1.py
@@ -8,8 +8,9 @@
 #
 
 from cgroup import Cgroup, CgroupVersion
-from run import Run, RunError
 from systemd import Systemd
+from process import Process
+from run import RunError
 import consts
 import ftests
 import time
@@ -91,12 +92,6 @@ def create_process_get_pid(config, CGNAME, SLICENAME='', ignore_systemd=False):
     return pids, result, cause
 
 
-def terminate_process(config, pids):
-    if pids:
-        for p in pids.splitlines():
-            Run.run(['sudo', 'kill', '-9', p])
-
-
 def test(config):
     global SYSTEMD_PIDS, OTHER_PIDS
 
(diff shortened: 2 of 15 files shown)


tests/ftests/049-sudo-systemd_create_scope.py
```diff
@@ -63,7 +63,7 @@ def teardown(config, result):
     pid = CgroupCli.get(config, cgname=os.path.join(SLICE, SCOPE), setting='cgroup.procs',
                         print_headers=False, values_only=True)
     if pid is not None:
-        Run.run(['kill', '-9', str(pid)], shell_bool=True)
```
drakenclimber, line 66: The `str()` cast isn't needed in this case because the PID came from `cgget` which returns it as a string already

tests/ftests/065-sudo-systemd_cgclassify-v1.py
```diff
@@ -93,8 +93,8 @@ def create_process_get_pid(config, CGNAME, SLICENAME='', ignore_systemd=False):
```
drakenclimber, line 93: I'm not 100% certain that I updated this test properly.  I did intentionally force errors and verify that the error paths behave mostly sanely, but it would be good to have another set of eyes look at this.

tests/ftests/066-sudo-systemd_cgclassify-v2.py
```diff
@@ -104,9 +104,13 @@ def create_process_get_pid(config, CGNAME, SLICENAME='', ignore_systemd=False):
```
drakenclimber, line 104: Same comment as the previous test.

tests/ftests/067-sudo-systemd_cgexec-v1.py
```diff
@@ -96,8 +96,8 @@ def create_process_get_pid(config, CGNAME, SLICENAME='', ignore_systemd=False):
```
drakenclimber, line 96: Ditto

kamalesh-babulal:
@drakenclimber The cleanup looks great.   Had two points to improve, do you think they make sense:
1. For completeness in `process.py::__save_child_pid`, let's convert the `pid` to `int`, while adding it to `children_pids`, other places with this patchset deal with `int`:
```
diff --git a/tests/ftests/process.py b/tests/ftests/process.py
index a1015e6..7d3e45f 100644
--- a/tests/ftests/process.py
+++ b/tests/ftests/process.py
@@ -77,7 +77,7 @@ class Process(object):
             pid = Run.run(cmd, shell_bool=True)
 
             for _pid in pid.splitlines():
-                self.children_pids.append(_pid)
+                self.children_pids.append(int(_pid))
 
             if pid.find('\n') >= 0:
                 # The second pid in the list contains the actual perl process
```
2. Given that, the loop to kill process is written a couple of times across  test cases, should it be moved into its own helper function in `cgroup.py`

kamalesh-babulal:
> Given that, the loop to kill process is written a couple of times across test cases, should it be moved into its own helper function in cgroup.py

I tried implementing the idea https://github.com/libcgroup/libcgroup/commit/838483655ddf5e800b0b421f9aa3c8d7caba26c4

drakenclimber:
Thanks for the additional cleanup, @kamalesh-babulal.  Good idea.

I put the helper function in a different location, `Process.kill()`, but I kept your logic as is.  Let me know your thoughts.  (Oh and I also rebased to pick up the synchronize github runners logic.)

Thanks!

kamalesh-babulal:
I thought about the placement of `kill` function in `process.py` initially and then voted against it, thinking it was an extension to `cgroups` like a wrapper but rethinking it makes sense to place it in `process.py`.  There are a couple of `flake8` warnings about unused imports, I have a delta patch that would clean it up https://github.com/libcgroup/libcgroup/commit/083d015bd1196931684c702d6ec1bf3c326cdc0e. Feel free to use it and it also has the hunk:
```
index a2297c6a648f..97aa1734028e 100644
--- a/tests/ftests/process.py
+++ b/tests/ftests/process.py
@@ -77,7 +77,7 @@ class Process(object):
             pid = Run.run(cmd, shell_bool=True)
 
             for _pid in pid.splitlines():
-                self.children_pids.append(_pid)
+                self.children_pids.append(int(_pid))
 
             if pid.find('\n') >= 0:
                 # The second pid in the list contains the actual perl process
```

do you think, the above hunk makes sense?

kamalesh-babulal (approved):

PR #309, opened 2023-02-24
wrapper: fix segfault in cgroup_add_controller()

kamalesh-babulal:
This patch series fixes a segfault in the `cgroup_add_controller()` APIs.
When `NULL` is passed in place of the controller name.  It also adds
test cases to the gunit fuzzer.


drakenclimber (approved):

drakenclimber:
Pushed to main and release-3.0.  Thanks, @kamalesh-babulal

PR #310, opened 2023-02-24
 [RFC] tests/ftests: synchronize between github runners

drakenclimber:
If a VM is shared between multiple github runners, there are chances of
each other stomping over other's run, if executed parallelly.  To avoid the
race between the runners, introduce lock file, that gets acquired (created)
when `ftest.sh` starts and gets removed by `ftest-nocontainer.sh`, this
ensures that both test cases are executed before other runners, that's
waiting for their chance to run.  A runner would wait for 10 minutes before
re-trying to run.  At the max, a runner would wait for 50 minutes (5 retries)
before giving up.

This is essentially v2 of pull request #301


kamalesh-babulal (approved):

drakenclimber:

PR #312, opened 2023-02-28
wrapper: fix segfault in cgroup_get_uid_gid()

kamalesh-babulal:
This patch series fixes a segfault in the `cgroup_get_uid_gid()` API,
that triggers when NULL is passed in place of the permissions.
It also adds test cases to the gunit fuzzer.


drakenclimber (approved):

drakenclimber:

PR #319, opened 2023-03-06
Add cgroup v2 support to cgroup_add_all_controllers()

drakenclimber:
This patchset does a few related things:
* Add cgroup v2 suppor to `cgroup_add_all_controllers()`.  For cgroup v1, `cgroup_add_all_controllers()` reads `/proc/cgroups`.  `/proc/cgroups` is meaningless in cgroup v2, so for v2 read the cgroup's `cgroup.controllers.file`
* A previous patch modified `cgroup_get_cgroup()` to read the `cgroup.subtree_control` file to get the enabled controllers.  This isn't correct as the subtree control file dictates which controllers are enabled for the children.  Instead read the cgroup's `cgroup.controllers` file
* Add python bindings for `cgroup_add_all_controllers`
* And of course add tests for `cgroup_add_all_controllers()` for both v1 and v2.  The v1 behavior is slightly unintuitive, but I left it unchanged from earlier releases.  The v2 behavior means that a user can interact with libcgroup's APIs in multiple ways and will see consistent behavior.  See the below code snippet for an example of this

This patchset makes the following code equivalent:

```
#!/usr/bin/env python3
cg1 = Cgroup('user.slice', Version.CGROUP_V2)
cg1.get()

cg2 = Cgroup('user.slice', Version.CGROUP_V2)
cg2.add_all_controllers()
cg2.get()

### cg1 and cg2 will have the same list of controllers
```

Code coverage increased ~2.5% :)


src/wrapper.c
```diff
@@ -112,47 +112,107 @@ int cgroup_add_all_controllers(struct cgroup *cgroup)
...
+	if (!cgroup)
+		return ECGINVAL;
+
+	mode = cgroup_setup_mode();
+
+	/*
+	 * Per kernel documentation, cgroup-v2.rst, /proc/cgroups is "meaningless" for cgroup v2.
+	 * Use the cgroup's cgroup.controllers file instead
+	 */
+	if (mode == CGROUP_MODE_UNIFIED) {
```
drakenclimber, line 128: Github has mangled this diff.  I added the `if (mode == CGROUP_MODE_UNIFIED)` section, and the `else` section remained unchanged other than the indentation.

tests/ftests/051-sudo-cgroup_get_cgroup.py
```diff
@@ -14,7 +14,7 @@
 import sys
 import os
 
-CGNAME = '051cgnewcg'
```
drakenclimber, line 17: We need to use a child cgroup here because we don't know what controllers will be enabled in the root's `cgroup.subtree_control` file and this could vary from one system to another.  Using a child cgroup lets us explicitly dictate the enabled controllers.

tests/ftests/072-pybindings-cgroup_get_cgroup.py
```diff
@@ -101,7 +102,11 @@ def test(config):
     # Test 5 - Create a parent/child cgroup with no controllers enabled.  Ensure the user can get
     # the cgroup with no errors.  .get() should populate zero cgroups
     #
+    for ctrl in CONTROLLERS:
```
drakenclimber, line 105: Explicitly disable all controllers in the grandchild.

@kamalesh-babulal  Hmmmm.... should we make this a helper function in the `Cgroup` class?

tests/ftests/073-pybindings-cgroup_add_all_controllers-v2.py
```diff
@@ -0,0 +1,103 @@
...
+CONTROLLERS = ['cpu', 'memory', 'pids']
+SUBTREE_CONTROL = 'cpu'
+
+def prereqs(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    if config.args.container:
+        result = consts.TEST_SKIPPED
+        cause = 'This test cannot be run within a container'
```
drakenclimber, line 29: This should probably work in a container, but I didn't try very hard to get it working.  Thoughts, @kamalesh-babulal ?

src/wrapper.c
```diff
@@ -112,47 +112,107 @@ int cgroup_add_all_controllers(struct cgroup *cgroup)
...
 
-		/* add mounted controller to cgroup structure */
-		cgc = cgroup_add_controller(cgroup, info.name);
-		if (!cgc) {
-			ret = ECGINVAL;
-			fprintf(stderr, "controller %s can't be added\n", info.name);
-			goto end;
+		ret_c = fgets(line, CGV2_CONTROLLERS_LL_MAX, fp);
+		if (ret_c == NULL) {
+			/* no controllers are enabled */
```
kamalesh-babulal, line 155: @drakenclimber Patchset looks good.  I had a few minor comments on the patches.
- On no controllers available, bailout to `goto out` is missing.
- `fp` would leak in both successful/error paths. `fclose(fp)` can be done right after `fgets`

src/wrapper.c
```diff
@@ -112,47 +112,107 @@ int cgroup_add_all_controllers(struct cgroup *cgroup)
...
+	 * Per kernel documentation, cgroup-v2.rst, /proc/cgroups is "meaningless" for cgroup v2.
+	 * Use the cgroup's cgroup.controllers file instead
+	 */
+	if (mode == CGROUP_MODE_UNIFIED) {
+		/*
+		 * cg_cgroup_v2_mount_path (FILENAME_MAX) + cgroup->name (FILENAME_MAX) +
+		 * strlen("cgroup.controllers") (18) + 2 forward slashes + 1 NULL terminator
+		 */
+		char cgroup_controllers_path[FILENAME_MAX * 2 + 18 + 2 + 1];
+		char *ret_c, *controller, *stok_buff = NULL, line[CGV2_CONTROLLERS_LL_MAX];
```
kamalesh-babulal, line 134: Minor nitpick, can yo please follow the reverse xmas tree declaration.

src/api.c
```diff
@@ -3628,8 +3667,8 @@ int cgroup_get_cgroup(struct cgroup *cgroup)
 		} else { /* cgroup v2 */
 			bool enabled;
 
-			error = cgroupv2_get_subtree_control(path, cg_mount_table[i].name,
-							     &enabled);
+			error = cgroupv2_get_controllers(path, cg_mount_table[i].name,
+							 &enabled);
```
kamalesh-babulal, line 3671: It should be `cgroupv2_get_subtree_control()`? or else the `cg.get()` will return `cgroup.controllers` controllers.
```
>>> cg = Cgroup("a", Version.CGROUP_V2)
>>> cg.get()
>>> len(cg.controllers)
5
```
replacing `cgroupv2_get_controllers()` with `cgroupv2_get_subtree_control()`:
```
>>> from libcgroup import Cgroup, Version
>>> cg = Cgroup("a", Version.CGROUP_V2)
>>> cg.get()
>>> len(cg.controllers)
1
```

tests/ftests/072-pybindings-cgroup_get_cgroup.py
kamalesh-babulal: Yeah, we could, its will be useful at other places too.

tests/ftests/073-pybindings-cgroup_add_all_controllers-v2.py
kamalesh-babulal: It should work in the container too, let me try it out in the container too ensure there are no issues.

tests/ftests/073-pybindings-cgroup_add_all_controllers-v2.py
```diff
@@ -0,0 +1,104 @@
...
+        return result, cause
+
+    if CgroupCli.get_cgroup_mode(config) != Mode.CGROUP_MODE_UNIFIED:
+        result = consts.TEST_SKIPPED
+        cause = 'This test requires the unified cgroup hierarchy'
+        return result, cause
```
kamalesh-babulal, line 36: `return` here can be removed, as the next statement is a `return`

tests/ftests/073-pybindings-cgroup_add_all_controllers-v2.py
```diff
@@ -0,0 +1,104 @@
...
+def setup(config):
+    CgroupCli.create(config, CONTROLLERS, PARENTCG)
+
+    # Ensure that the subtree control file differs from the cgroup.controllers file
+    for ctrl in CONTROLLERS:
+        CgroupCli.set(config, PARENTCG, 'cgroup.subtree_control',
+                      '-{}'.format(ctrl), ignore_systemd=True)
+    CgroupCli.set(config, PARENTCG, 'cgroup.subtree_control',
+                  '+{}'.format(CONTROLLERS[0]), ignore_systemd=True)
+    CgroupCli.get_and_validate(config, PARENTCG, 'cgroup.subtree_control', CONTROLLERS[0], True)
```
kamalesh-babulal, line 50: I guess `SUBTREE_CONTROL` was intended to be used in place of `CONTROLLERS[0]`.

tests/ftests/074-pybindings-cgroup_add_all_controllers-v1.py
```diff
@@ -0,0 +1,91 @@
...
+
+def prereqs(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    if CgroupCli.get_cgroup_mode(config) != Mode.CGROUP_MODE_LEGACY:
+        result = consts.TEST_SKIPPED
+        cause = 'This test requires the legacy cgroup hierarchy'
+        return result, cause
+
```
kamalesh-babulal, line 29: The `return` can be removed, the next statement is a `return` too.

tests/ftests/074-pybindings-cgroup_add_all_controllers-v1.py
```diff
@@ -0,0 +1,91 @@
...
+def setup(config):
+    CgroupCli.create(config, CONTROLLERS, CGNAME)
+
+
+def test(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    cgall = Cgroup(CGNAME, Version.CGROUP_V2)
```
kamalesh-babulal, line 41: This should have been `Version.CGROUP_V1`

tests/ftests/072-pybindings-cgroup_get_cgroup.py
```diff
@@ -14,10 +14,11 @@
 import sys
 import os
 
-CGNAME = '072cggetcg'
+
```
kamalesh-babulal, line 17: This fixes the warning `E302 expected 2 blank lines, found 1` :)

src/api.c
drakenclimber: My position is that reading the `subtree_control` file is incorrect in this case.  The subtree control file tells us what controllers are enabled for the _children_ cgroups, but it has no bearing on the controllers enabled in _this_ cgroup.  To get the controllers enabled in this cgroup we should be reading the `cgroup.controllers` file.

src/wrapper.c
drakenclimber: Oops.  Good catches

src/api.c
kamalesh-babulal: Sorry, I missed a detail in the comment, I had enabled `cpu` controller using `# echo +cpu > /sys/fs/cgroup/a/cgroup.subtree_control`.  I am a little confused, I thought `cgroup.controllers` provides the list of controllers that can be enabled for the cgroup vs `cgroup.subtree_control` is the list of controllers enabled from the available controllers listed in `cgroup.controllers`.

drakenclimber: Or here's another way to think about it:
```
# mkdir -p /sys/fs/cgroup/foo/bar
# ls /sys/fs/cgroup/foo/bar/ | wc -l
15

### At this point, there are no enabled controllers in foo/bar

# echo +cpu > /sys/fs/cgroup/foo/cgroup.subtree_control
# echo +memory > /sys/fs/cgroup/foo/cgroup.subtree_control
# ls /sys/fs/cgroup/foo/bar/ | wc -l
36
# cat /sys/fs/cgroup/foo/bar/cgroup.subtree_control

### Now there are a whole bunch more files in foo/bar because the cpu
### and memory controllers have been enabled within bar.   And foo/bar's
### cgroup.subtree_control is empty, so if we were to make foo/bar/baz,
### it would have no activated controllers

# mkdir /sys/fs/cgroup/foo/bar/baz
# ls /sys/fs/cgroup/foo/bar/baz | wc -l
15
```

kamalesh-babulal: The answer is way more complicated than one may expect, when a cgroup is created using `mkdir`, they need to know what resources/controllers the tasks in that cgroup will be restricted too, defining the characteristics of the newly created cgroup. Hence, `cgroup.controllers` file will always be an intersection of `cgroup->parent.subtree_control` file, defining the controllers available for the cgroup. One may argue, If that the case why are the knobs of the controller those are not enabled by writing into `cgroup.subtree_control` is visible along with the `cgroup.*` files such `cgroup.procs`, and they should be only visible when the controller is enabled? To answer that, we need to look at the other constraint that cgroup v2 imposes, which is only the leaf node without
any controller enabled should have the tasks. This can be better illustrated
with a diagram:
```
                  /
                c1
                 |
               c2
               /  \
             t1 t2
```

before we dive into the example, the crux of the cgroup in one-liner is "what controller resource is restricted vs unrestricted", all the resources are unrestricted until the controller is enabled in the parent cgroup or we explicitly
turn on some controller settings in the cgroup the tasks are running.

With the above example, `c1` is the parent, that decides which all controller resources shall be restricted to the task `t1` and `t2` by enabling and disabling the controller in its `cgroup.subtree_control` file.  Coming to our issue here, when we try to find the controllers enabled vs available or in other words unrestricted vs restricted to the tasks or children cgroups, we should be reading the `cgroup.subtree_control` of the parent or read `cgroup.controllers` file instead to find the controller who is restricting/enabled. Hence `cgroupv2_get_controllers()` seems to be the right option to go with.

kamalesh-babulal (approved):

drakenclimber:
Pushed to the main and release-3.0 branches.  Thanks for all your help, @kamalesh-babulal

PR #321, opened 2023-03-08
ftests: renumber the test case 098 to 998

kamalesh-babulal:
The test cases, that cover the unlikely conditions, are numbered in
the reverse starting from `999`, similarly, test case `098` tries to
delete a non-existent shared point.  Re-number it from `099` -> `998`
to hint at the nature of the test case.


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 259928ed2aaa.  Pushed to release-3.0 in commit 751d25b6839b.  Thanks, @kamalesh-babulal

PR #324, opened 2023-03-13
[RFC] cgconfig: auto convert cgroup controller versions

kamalesh-babulal:
The user might be using cgroup v1 controller settings in the cgconfig.conf and when migrating to the cgroup v2 setup, they are required to manually edit the cgconfig.conf file to map v1 controllers settings to that of v2 settings, this might be cucumber activity.  This patch tries to address the mapping issue, by silently mapping the possible controller settings to the current cgroup setup.

Consider the scenario:
```
$ cgget -m
Unified Mode (Cgroup v2 only).

$ cat cgconfig.conf
group foo {
        cpu {
                cpu.shares = 512;
                cpu.cfs_quota_us = 5000;
        }
        cpuset {
                cpuset.cpus = 0-1;
        }
}
```
the cpu controller settings are of cgroup v1 but the system is booted into cgroup v2 setup. Running cgconfigparser without the patch, the user will be warned about invalid settings and the parser will exit.
```
$ sudo cgconfigparser -l ./cgconfig.conf
cgconfigparser; error loading cgconfig.conf: Cgroup, requested group parameter does not exist

with this patch, cgconfigparser will attempt to convert the settings. $ sudo cgconfigparser -l ./cgconfig.conf
$ cat /sys/fs/cgroup/foo/cpu.weight
50
$ cat /sys/fs/cgroup/foo/cpu.max
5000 100000
```


src/config.c
```diff
@@ -754,6 +754,104 @@ static int cgroup_config_mount_fs(void)
...
+		return ECGROUPNOTCREATED;
+	}
+
+	/* Copy the non-controllers details */
+	cgrp_cpy->index = 0;
+	cgrp_cpy->task_fperm = cgroup->task_fperm;
+	cgrp_cpy->control_fperm = cgroup->control_fperm;
+	cgrp_cpy->control_dperm = cgroup->control_dperm;
+
+	cgrp_cpy->control_gid = cgroup->control_gid;
```
drakenclimber, line 783: Would it make sense to call `cgroup_set_uid_gid()` here?

src/config.c
```diff
@@ -754,6 +754,104 @@ static int cgroup_config_mount_fs(void)
...
+
+	cgrp_cpy = cgroup_new_cgroup(cgroup->name);
+	if (!cgrp_cpy) {
+		cgroup_err("Failed to create cgroup %s\n", cgroup->name);
+		return ECGROUPNOTCREATED;
+	}
+
+	/* Copy the non-controllers details */
+	cgrp_cpy->index = 0;
+	cgrp_cpy->task_fperm = cgroup->task_fperm;
```
drakenclimber, line 779: Would it make sense to call `cgroup_set_permissions()` here?

src/config.c
```diff
@@ -767,8 +865,15 @@ static int cgroup_config_create_groups(void)
...
 		error = cgroup_create_cgroup(cgroup, 0);
 		cgroup_dbg("creating group %s, error %d\n", cgroup->name, error);
-		if (error)
-			return error;
+		if (error) {
+			/*
+			 * Attempt to convert the controller version
+			 * and retry.
+			 */
+			error = convert_controller_versions(cgroup);
```
drakenclimber, line 873: How would this work on a hybrid system where some controllers are v1 and some are v2?   And do we care?

src/config.c
```diff
@@ -767,8 +865,15 @@ static int cgroup_config_create_groups(void)
 
 		error = cgroup_create_cgroup(cgroup, 0);
```
drakenclimber, line 866: Would it make more sense to try and convert the cgroup prior to calling `cgroup_create_cgroup()`.  It feels weird to try and write it, watch it fail, then convert it and try again.

What if we added a flag to the cgconfig file like this:
```
group foo {
        version = 1;
        cpu {
                cpu.shares = 512;
                cpu.cfs_quota_us = 5000;
        }
        cpuset {
                cpuset.cpus = 0-1;
        }
}
```

Then the convert logic would look something like:
```
mode = get_mode();
/// do some trickery to map mode to cgroup version
/// if mode == hybrid, probably skip the conversion because hybrid is weird and hard to work with

if system_cg_version != config_cg_version:
    cgroup_convert_cgroup();

cgroup_create_cgroup()
```

drakenclimber: Actually we could probably handle the hybrid case as well if we really wanted to.  We might have to split up the cgroup if it has different target versions (say cpu controller is v1 and memory controller is v2).  Or another idea would be to expose `convert_controller()` outside of abstraction-common.c

src/config.c
kamalesh-babulal: @drakenclimber Thank you so much for the review and apologies for the late response.  Make sense to call `cgroup_set_uid_gid()`.

src/config.c
kamalesh-babulal: Agreed, will use `cgroup_set_permissions()`.

src/config.c
kamalesh-babulal: It works for all the cases, please consider a hybrid cgroup setup, where the `cpu` controller is mounted as cgroup v2 controller:
```
# cat /proc/cmdline 
BOOT_IMAGE=/vmlinuz-6.2.0-rc2+ root=/dev/mapper/ubuntu--vg-ubuntu--lv ro maybe-ubiquity console=ttyS0,9600 console=ttyS1,9600 systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller=false cgroup_no_v1=cpu
# cgget -m
Hybrid mode (Cgroup v1/v2).

# cat cgconfig.conf
group a {
        cpu {
                cpu.shares = 756;
        }
}
```
```
# ./src/tools/cgget -r cpu.weight foo
foo:
cpu.weight: 100
```
```
# ./src/tools/cgconfigparser -l ./cgconfig.conf 
# ./src/tools/cgget -r cpu.weight foo
foo:
cpu.weight: 73
```

src/config.c
kamalesh-babulal: > Would it make more sense to try and convert the cgroup prior to calling cgroup_create_cgroup(). It feels weird to try and write it, watch it fail, then convert it and try again.

It's an interesting idea to use hints.  I guess the commit explanation is horrible in doing its job of explaining the approach, please allow me to explain the idea again and let me know, your thoughts on the approach, consider following the `cgconfig.conf`
```
group foo {
   cpuset {
           cpuset.cpus = 0;
   }
   cpu {
           cpu.shares = 512;
   }
}
```
parsing of the above file by the `cgconfigparser` will fail on cgroup v2 setup, since `cpu.shares` is cgroup v1 `cpu` controller setting.  It is easy to find the incorrect settings by glancing over the `cgconfig.conf` file but it gets difficult to guess the controller with incorrect settings during creation of cgroup `foo` with `cpu, cpuset` controllers from `cgconfigparser` code path.  The crux of the approach is: 
(for the sake of readability, `src_cgrp`  refers to the cgroup created by the `configparser` and `dst_cgrp`  refers to the cgroup created by the conversion code.)
1. Create `dst_cgrp` and  copy the permissions from the `src_cgrp`
2. Loop over `src_cgrp->controller[index]` and attach the `ith` controller (attach only one controller)
3.  Try creating the `dst_cgrp` with only `ith` controller from `src_cgrp`, and if it succeeds, move on to the next controller.  With the above example, `src_dst->index == 0` is `cpuset` controller, and `dst_cgrp` is created with only `cpuset` controller succeeds.
4. If step 3 succeeds,  the `dst_cgrp->controller`  index is reset to zero, (resetting `dst_cgrp` to empty cgroup with no controllers) and the single allocated controller is freed.  In our example, it's the  `cpuset` controller, which gets free'd, and repeat the same for the `i+1` controller of `src_cgrp` by starting from step 2.
5. When step 3  fails, conversion is attempted for the `dst_cgrp`,  with the controllers `convert_version` derived from simple boolean logic, i.e, if the current controller version is cgroup v1, then cgroup v2 is the `convert_version` and vice versa.  The logic is independent of checking the current cgroup setup mode to derive the `convert_version`.
6. If step 5 succeeds, repeat the steps from step 2 for the `i+1` controller or execute the error path on failing to convert the `dst_cgrp` controller setting to `convert_version`.

previously any one incorrect setting of a controller in the `cgconfig.conf`, will fail to create the cgroup.  But in the above patch,  the cgroup might exist with other controllers other than the first controller with incorrect settings, depending upon the order of controllers as parsed by the `cgconfigparser`.  If the above `cgconfig.conf` is modified to:
```
group foo {
   cpuset {
           cpuset.cpus = 0;
   }
   cpu {
           cpu.unknown = 512;
   }
}
```
in the above example, cgroup `foo` will exist with `cpuset` controller enabled and will execute the error path on the failed conversion of `cpu.unknown` setting of the `cpu` controller.  Such failure can be handled by either falling back to the original behavior of not creating the cgroup when an incorrect setting is encountered or skipping the faulty controller and creating the cgroup with the good controllers.  In the above patch, neither of the approaches is used and it needs to be fixed.

kamalesh-babulal: Actually returning the nonzero will call `cgroup_config_destroy_groups()` in the error path, cleaning up the cgroups created by the `cgconfig.conf` on the first occurrence of error/wrong/unconvertable cgroup controller settings.

kamalesh-babulal:
**V2 changes:**
- Used `cgroup_set_uid_gid()` and `cgroup_set_permissions()` as recommended by @drakenclimber 
- Added test case to test the auto-convert functionality.

kamalesh-babulal:
**v3 Changes:**
- Renamed the test case from `076` to `078`, to avoid conflict with PR's.

src/config.c
```diff
@@ -754,6 +754,103 @@ static int cgroup_config_mount_fs(void)
...
+		 * let try to convert the controller to the current version
+		 * and retry.
+		 */
+		convert_cgrp = cgroup_new_cgroup(cgroup->name);
+		if (!convert_cgrp) {
+			cgroup_err("Failed to created cgroup %s\n", cgroup->name);
+			ret = ECGROUPNOTCREATED;
+			goto err;
+		}
+		ret = cgroup_convert_cgroup(convert_cgrp, current_version, cgrp_cpy,
```
drakenclimber, line 831: Help me out here.  I think this is a "me" problem :)

Here's the simplified prototype for convert:
```cgroup_convert_cgroup(out_cg, out_version, in_cg, in_version)```

So doesn't that mean that we have the versions backward on line 831?  I see that the test passed, so it seems to be working.  But I don't understand how.

File: src/config.c
Comment (drakenclimber) [inline src/config.c:1]: I think "cucumber" in the commit comment above may not be the word we're looking for :) :)

Comment (kamalesh-babulal) [reply src/config.c:1]: @drakenclimber the intended word was `cumbersome`, apologises, and will fix it in the next version.

File: src/config.c (hunk 5 above)
Comment (kamalesh-babulal) [reply src/config.c:831]: Great catch, the issue is with the wrong variables used.
```
diff --git a/src/config.c b/src/config.c
index 8d59aca..c5c8bf9 100644
--- a/src/config.c
+++ b/src/config.c
@@ -786,8 +786,8 @@ static int convert_controller_versions(struct cgroup *cgroup)
                               cgroup->control_fperm, cgroup->task_fperm);
 
        for (i = 0; i < cgroup->index; i++) {
-               current_version = cgroup->controller[i]->version;
-               convert_version = current_version == CGROUP_V1 ? CGROUP_V2 : CGROUP_V1;
+               convert_version = cgroup->controller[i]->version;
+               current_version = convert_version == CGROUP_V1 ? CGROUP_V2 : CGROUP_V1;
 
                cgc = cgroup_add_controller(cgrp_cpy, cgroup->controller[i]->name);
                if (!cgc) {
@@ -828,8 +828,8 @@ static int convert_controller_versions(struct cgroup *cgroup)
                        ret = ECGROUPNOTCREATED;
                        goto err;
                }
-               ret = cgroup_convert_cgroup(convert_cgrp, current_version, cgrp_cpy,
-                                           convert_version);
+               ret = cgroup_convert_cgroup(convert_cgrp, convert_version, cgrp_cpy,
+                                           current_version);
                if (ret) {
                        cgroup_err("Conversion of controller failed.\n");
```
will respin v4 with the comments `current_version` and `convert_version`

kamalesh-babulal:
**V4 Changes:**
- Exchanged the usage of `convert_version` and `current_version` variables in `convert_controller_versions()`

drakenclimber:

PR #325, opened 2023-03-21
cleanup cgroup_set_values_recursive() to write dirty values only

kamalesh-babulal:
`cgroup_set_values_recursive()` is called by `cgroup_modify_cgroup()` to
modify controller values, where all the settings modified or not are
written to the disk.  This breaks when writing a new value of a setting
that is linked to another setting of the controller, followed by writing
an unmodified value to the linked setting. This effectively undoes the
modification. For example, consider two linked settings of the cpu
controller: `cpu.weight` and `cpu.weight.nice`, where writing the new value
of `cpu.weight` is followed by unmodified `cpu.weight.nice` value. Writing
of the latter will undo the new value of the former.

This three-patch, series cleans up the `cgroup_set_values_recursive()` by
renaming the third argument `ignore_non_dirty_failure` to
`ignore_non_dirty_values`.  This renames also changes the purpose of the
flag, where the calling functions, set it to ignore the writing of the
controller setting, which is not modified/dirty and introduces extensive
checks for writing the controller setting.  It also, sets the dirty flag unconditionally,
for destination controller values in cgroup_copy_controller_values() and
propagate the changes to the gunit test case.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -2287,41 +2287,73 @@ static int cg_set_control_value(char *path, const char *val)
  *
  * @param base The full path to the base of this cgroup
  * @param controller The controller whose values are being updated
+ * @param ignore_non_dirty_values If set skips writing non-dirty controller settings
  */
 STATIC int cgroup_set_values_recursive(const char * const base,
 				       const struct cgroup_controller * const controller,
-				       bool ignore_non_dirty_failures)
+				       bool ignore_non_dirty_values)
 {
+	struct control_value *cv;
+	struct stat path_stat;
 	int ret, j, error = 0;
 	char *path = NULL;
 
 	for (j = 0; j < controller->index; j++) {
-		ret = asprintf(&path, "%s%s", base, controller->values[j]->name);
+		cv = controller->values[j];
+
+		/*
+		 * ignore_non_dirty_values is set while writing into
+		 * existing cgroup to modify controller settings and
+		 * unset during cgroup creation.  The subtle difference
+		 * is that dirty flag is unset for all the controller
+		 * settings during cgroup creation, whereas some or all
+		 * controller settings have the dirty flag set during
+		 * modification.
+		 *
+		 * Be careful with ignore_non_dirty_values flag, setting
+		 * it writing only the controller settings that has it
+		 * dirty value set.
+		 */
+		if (ignore_non_dirty_values && cv->dirty == false)
+			continue;
+
+		/* We don't support, writing multiline settings */
+		if (strcspn(cv->value, "\n")  < (strlen(cv->value) - 1))
+			continue;
+
+		ret = asprintf(&path, "%s%s", base, cv->name);
 		if (ret < 0) {
 			last_errno = errno;
 			error = ECGOTHER;
 			goto err;
 		}
-		cgroup_dbg("setting %s to \"%s\", pathlen %d\n", path,
-			   controller->values[j]->value, ret);
 
-		error = cg_set_control_value(path, controller->values[j]->value);
+		/* skip read-only settings */
+		ret = (stat(path, &path_stat));
+		if (ret < 0) {
+			last_errno = errno;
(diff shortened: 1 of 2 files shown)


src/api.c
```diff
@@ -2287,41 +2287,73 @@ static int cg_set_control_value(char *path, const char *val)
...
+		cgroup_dbg("setting %s to \"%s\", pathlen %d\n", path, cv->value, ret);
 
-		controller->values[j]->dirty = false;
+		error = cg_set_control_value(path, cv->value);
+		free(path);
+		path = NULL;
+
+		if (error) {
+			/* Ignore the errors on depreciated settings */
+			if (last_errno == EOPNOTSUPP) {
```
drakenclimber, line 2350: I know `EOPNOTSUPP` is used for threaded cgroups.  Are there other use cases that return `EOPNOTSUPP`?

src/api.c
```diff
@@ -2287,41 +2287,73 @@ static int cg_set_control_value(char *path, const char *val)
...
-			goto err;
+		cgroup_dbg("setting %s to \"%s\", pathlen %d\n", path, cv->value, ret);
 
-		controller->values[j]->dirty = false;
+		error = cg_set_control_value(path, cv->value);
+		free(path);
+		path = NULL;
+
+		if (error) {
+			/* Ignore the errors on depreciated settings */
```
drakenclimber, line 2349: silly nitpick - deprecated not depreciated

drakenclimber (approved):
I had a couple silly comments, but I think this is a good improvement, @kamalesh-babulal.  Thanks!!

src/api.c
kamalesh-babulal: I have observed one such instance in `unified` hirerachy:
```
# CGROUP_LOGLEVEL=DEBUG cgcreate -gcpuset,cpu,cpuacct,blkio,memory,devices,freezer,net_cls,perf_event,net_prio,hugetlb,pids,rdma,misc:foo
# CGROUP_LOGLEVEL=DEBUG cgcreate -gcpuset,cpu,cpuacct,blkio,memory,devices,freezer,net_cls,perf_event,net_prio,hugetlb,pids,rdma,misc:bar
# CGROUP_LOGLEVEL=DEBUG ./src/tools/cgset --copy-from foo bar &> /tmp/e
# cat /tmp/e|grep -i return
/sys/fs/cgroup/memory//bar/memory.kmem.limit_in_bytes returned EOPNOTSUPP
```

I had the following hacky debug patch over this patchset:
```
--- a/src/api.c
+++ b/src/api.c
@@ -2342,17 +2342,23 @@ STATIC int cgroup_set_values_recursive(const char * const base,
                cgroup_dbg("setting %s to \"%s\", pathlen %d\n", path, cv->value, ret);
 
                error = cg_set_control_value(path, cv->value);
-               free(path);
-               path = NULL;
 
                if (error) {
                        /* Ignore the errors on depreciated settings */
                        if (last_errno == EOPNOTSUPP) {
+                               cgroup_dbg("%s returned EOPNOTSUPP\n", path);
                                error = 0;
+                               free(path);
+                               path = NULL;
                                continue;
                        }
+                       cgroup_dbg("%s returned error code %d\n", path, error);
+                       free(path);
+                       path = NULL;
                        goto err;
                }
+               free(path);
+               path = NULL;
                cv->dirty = false;
        }
 
```

src/api.c
kamalesh-babulal: :), the intention was to use the non financial terms, will fix it in the v2

kamalesh-babulal:
@drakenclimber Thank you so much for the review.

drakenclimber:
Pushed to the main and release-3.0 branches.  Thanks, @kamalesh-babulal and @justinfx.

src/api.c
drakenclimber: No worries :).  I fixed the typo in the push to main

PR #328, opened 2023-04-06
Add python bindings and a test for cgroup_get_procs()

drakenclimber:


tests/ftests/076-pybindings-cgroup_get_procs.py
```diff
@@ -0,0 +1,112 @@
...
+    global initial_pid_list
+    CgroupCli.create(config, CONTROLLERS, CGNAME)
+    CgroupCli.create(config, CONTROLLERS, EMPTY_CGNAME)
+
+    for i in range(0, PID_CNT):
+        pid = config.process.create_process(config)
+        initial_pid_list.append(pid)
+
+    CgroupCli.classify(config, CONTROLLERS, CGNAME, initial_pid_list, ignore_systemd=True)
+    initial_pid_list = initial_pid_list.sort()
```
kamalesh-babulal, line 49: @drakenclimber patch looks good. A few minor nitpicks, `cgroup_get_procs()` does a sort before returning the pid list, do you see an unsorted pid list?

tests/ftests/076-pybindings-cgroup_get_procs.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def prereqs(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    if config.args.container:
+        result = consts.TEST_SKIPPED
+        cause = 'This test cannot be run within a container'
+        return result, cause
```
kamalesh-babulal, line 34: The next statement is a `return` too, hence this statement can be removed.

tests/ftests/076-pybindings-cgroup_get_procs.py
drakenclimber: I have seen the list be in a different order coming back from `cgroup_get_procs()`, and this caused failures as I was writing the test.  Not sure why it was out of order, honestly.

It's likely not needed after calling `create_process()` as the PIDs coming back from it should be monotonically increasing, but I don't see a downside to doing it.

kamalesh-babulal: Agreed, there is not down side to sorting it. I was just curious.

kamalesh-babulal (approved):

drakenclimber:

PR #329, opened 2023-04-06
Add systemd support to cgcreate

drakenclimber:
Add support for creating systemd scopes using the CLI tool cgcreate.

### DIFF
--- a/include/libcgroup/systemd.h
+++ b/include/libcgroup/systemd.h
@@ -119,6 +119,18 @@ int cgroup_add_systemd_opts(const char * const conf, const char * const value);
  */
 void cgroup_cleanup_systemd_opts(void);
 
+/*
+ * Write the specified slice and scope to the libcgroup systemd run file. This
+ * slice and scope will then be used as the default cgroup root. Subsequent
+ * libcgroup commands, cgget, etc., will utilize this slice and scope when
+ * constructing the libcgroup path
+ *
+ * @param slice Slice name, e.g. libcgroup.slice
+ * @param scope Scope name, e.g. database.scope
+ */
+int cgroup_write_systemd_default_cgroup(const char * const slice,
+				        const char * const scope);
+
 #ifdef __cplusplus
 } /* extern "C" */
 #endif
--- a/src/systemd.c
+++ b/src/systemd.c
@@ -88,12 +88,12 @@ int cgroup_create_scope(const char * const scope_name, const char * const slice_
 			const struct cgroup_systemd_scope_opts * const opts)
 {
 	sd_bus_message *msg = NULL, *reply = NULL;
+	int ret = 0, sdret = 0, cgret = ECGFAIL;
 	sd_bus_error error = SD_BUS_ERROR_NULL;
 	const char *job_path = NULL;
 	struct timespec start, now;
 	sd_bus *bus = NULL;
 	pid_t child_pid;
-	int ret = 0;
 
 	if (!scope_name || !slice_name || !opts)
 		return ECGINVAL;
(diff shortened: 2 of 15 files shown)


drakenclimber:
The merge conflict is due to this patchset being based upon pull #328.  Automated tests passed here:
https://github.com/drakenclimber/libcgroup/actions/runs/4632445865

Code coverage went up 0.2% to 56.8%
https://coveralls.io/builds/58649161

src/systemd.c
```diff
@@ -281,7 +296,7 @@ int cgroup_create_scope(const char * const scope_name, const char * const slice_
 	sd_bus_message_unref(reply);
 	sd_bus_unref(bus);
 
-	return ret;
+	return cgret;
```
kamalesh-babulal, line 299: `cgret` can be initialized to `ECGFAIL` and set to `cgret = 0`, instead of `ret = 0` in line 288, it would save some lines, it is a coding style comment, please feel free to ignore it.

src/tools/cgcreate.c
```diff
@@ -221,7 +226,19 @@ int main(int argc, char *argv[])
...
+
+			/*
+			 * Intentionally ignoring the return code because we are providing a
+			 * valid pointer
+			 */
+			(void)cgroup_set_default_scope_opts(&opts);
+
+			ret = cgroup_create_scope2(cgroup, 0, &opts);
+		} else {
+			ret = cgroup_create_cgroup(cgroup, 0);
```
kamalesh-babulal, line 240: When creating a delegated slice and scope, the `<slice>.name/<scope>.scope` should be written to `/var/run/libcgroup/systemd`, otherwise the delegated slice/scope will not be appended to the default cgroup path. Consider the following example:
```
# ./src/tools/cgcreate -c -g cpu:foo.slice/bar.scope
# ./src/tools/cgset -r cpu.weight=50 /
/root/libcgroup/src/tools/.libs/cgset: cgroup modify error: Cgroup, requested group parameter does not exist
```
Some like:
```
diff --git a/include/libcgroup/systemd.h b/include/libcgroup/systemd.h
index 39f6974..9a57576 100644
--- a/include/libcgroup/systemd.h
+++ b/include/libcgroup/systemd.h
@@ -119,6 +119,8 @@ int cgroup_add_systemd_opts(const char * const conf, const char * const value);
  */
 void cgroup_cleanup_systemd_opts(void);
 
+int cgroup_write_systemd_default_cgroup(const char * const slice,
+                                       const char * const scope);
 #ifdef __cplusplus
 } /* extern "C" */
 #endif
diff --git a/src/config.c b/src/config.c
index 6b4fe06..be50420 100644
--- a/src/config.c
+++ b/src/config.c
@@ -1968,7 +1968,7 @@ static int remove_systemd_default_cgroup_file(void)
  * cgroup_set_default_systemd_cgroup() for setting
  * systemd_default_cgroup used to form the cgroup path.
  */
-static int cgroup_write_systemd_default_cgroup(const char * const slice,
+int cgroup_write_systemd_default_cgroup(const char * const slice,
                                               const char * const scope)
 {
        FILE *systemd_def_cgrp_f;
@@ -2172,4 +2172,8 @@ int cgroup_alloc_systemd_opts(const char * const config, const char * const valu
 void cgroup_cleanup_systemd_opts(void) { }
 
 void cgroup_set_default_systemd_cgroup(void) { }
+int cgroup_write_systemd_default_cgroup(const char * const slice,
+                                              const char * const scope) {
+       return 1;
+}
 #endif
diff --git a/src/libcgroup.map b/src/libcgroup.map
index 28ce7d4..b0f0356 100644
--- a/src/libcgroup.map
+++ b/src/libcgroup.map
@@ -156,4 +156,5 @@ CGROUP_3.0 {
        cgroup_get_controller_name;
        cgroup_create_scope2;
        cgroup_set_default_systemd_cgroup;
+       cgroup_write_systemd_default_cgroup;
 } CGROUP_2.0;
diff --git a/src/tools/cgcreate.c b/src/tools/cgcreate.c
index 8a46a8e..29f2417 100644
--- a/src/tools/cgcreate.c
+++ b/src/tools/cgcreate.c
@@ -228,6 +228,9 @@ int main(int argc, char *argv[])
 
                if (create_scope) {
                        struct cgroup_systemd_scope_opts opts;
+                       char slice[FILENAME_MAX];
+                       char *scope;
+                       int len;
 
                        /*
                         * Intentionally ignoring the return code because we are providing a
@@ -236,6 +239,22 @@ int main(int argc, char *argv[])
                        (void)cgroup_set_default_scope_opts(&opts);
 
                        ret = cgroup_create_scope2(cgroup, 0, &opts);
+                       if (!ret) {
+                               scope = strstr(cgroup->name, "/");
+                               len = strlen(cgroup->name) - strlen(scope);
+                               strncpy(slice, cgroup->name, len);
+                               slice[len] = '\0';
+                               scope++;
+                               ret = cgroup_write_systemd_default_cgroup(slice, scope);
+                               if (!ret) {
+                                       err("%s: failed to write default %s/%s to ",
+                                               argv[0], slice, scope);
+                                       err("/var/run/libcgroup/systemd\n");
+                                       ret = 1;
+                                       goto err;
+                               }
+                               ret = 0;
+                       }
                } else {
                        ret = cgroup_create_cgroup(cgroup, 0);
                }

```
```
# ./src/tools/cgcreate -c -g cpu:foo.slice/bar.scope
# ./src/tools/cgset -r cpu.weight=50 /
# ./src/tools/cgget -r cpu.weight /
/:
cpu.weight: 50
```

or there can be a lightweight version of `cgroup_write_systemd_default_cgroup()`, implemented in the `cgcreate` tool, instead of exporting `cgroup_write_systemd_default_cgroup()`, that would effectively do:
- write `slice/scope` to `/var/run/libcgroup/systemd`
- skip the check to check if the cgroup (slice/scope) exists, because the `cgroup_create_scope2()` successful return means, the cgroup exists.

src/tools/cgcreate.c
```diff
@@ -169,8 +174,8 @@ int main(int argc, char *argv[])
 		goto err;
 	}
 
-	/* this is false always for disable-systemd */
-	if (!ignore_default_systemd_delegate_slice)
+	/* this will always be false if WITH_SYSTEMD is not defined */
+	if (!create_scope && !ignore_default_systemd_delegate_slice)
```
kamalesh-babulal, line 178: Both `create_scope` and `ignore_default_systemd_delegate_slice` are mutually exclusive, we need a check to check `T | T` premutation.

src/tools/cgcreate.c
drakenclimber: Good idea.  I'll play around with it.

src/tools/cgcreate.c
drakenclimber: I was thinking that as well.  Glad you brought it up.

src/tools/cgcreate.c
```diff
@@ -42,7 +42,8 @@ static void usage(int status, const char *program_name)
 	info("  -s, --tperm=mode		Tasks file permissions\n");
 	info("  -t <tuid>:<tgid>		Owner of the tasks file\n");
 #ifdef WITH_SYSTEMD
-	info("  -b				Ignore default systemd ");
+	info("  -b				Ignore default systemd\n");
+	info("  -c, --scope                     Create a delegated systemd scope ");
```
kamalesh-babulal, line 46: The indentation of `-c`, is not aligned, I guess is the tabs vs spaces issue.
```
# ./src/tools/cgcreate -h
[snip]
  -t <tuid>:<tgid>              Owner of the tasks file
  -b                            Ignore default systemd
  -c, --scope                     Create a delegated systemd scope delegate hierarchy
```

tests/ftests/077-sudo-cgreate_systemd_scope.py
```diff
@@ -0,0 +1,92 @@
...
+        pass
+
+
+def main(config):
+    [result, cause] = prereqs(config)
+    if result != consts.TEST_PASSED:
+        return [result, cause]
+
+    setup(config)
+    if result != consts.TEST_PASSED:
```
kamalesh-babulal, line 77: return from `setup()` is always `TEST_PASSED` and `result` here is return from `prereqs`, the return check can be removed.

tests/ftests/cgroup.py
```diff
@@ -163,6 +163,9 @@ def create(config, controller_list, cgname, user_name=None,
         if ignore_systemd:
             cmd.append('-b')
 
+        if create_scope:
+            cmd.append('-c')
```
kamalesh-babulal, line 167: The check for mutually exclusive flags `-b` and `-c` check needs to be done here, both can't be set at the same time.

kamalesh-babulal: We need to update `create_and_validate()` too.

tests/ftests/077-sudo-cgreate_systemd_scope.py
```diff
@@ -0,0 +1,92 @@
...
+
+def setup(config):
+    return consts.TEST_PASSED, None
+
+
+def test(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    Cgroup.create(config, CONTROLLERS, CGNAME, create_scope=True)
```
kamalesh-babulal, line 48: should we use the `create_and_validate()` variant here, to confirm the `slice/scope` was created under the right path.

include/libcgroup/systemd.h
```diff
@@ -119,6 +119,18 @@ int cgroup_add_systemd_opts(const char * const conf, const char * const value);
...
 void cgroup_cleanup_systemd_opts(void);
 
+/*
+ * Write the specified slice and scope to the libcgroup systemd run file. This
+ * slice and scope will then be used as the default cgroup root. Subsequent
+ * libcgroup commands, cgget, etc., will utilize this slice and scope when
+ * constructing the libcgroup path
+ *
+ * @param slice Slice name, e.g. libcgroup.slice
+ * #param scope Scope name, e.g. database.scope
```
kamalesh-babulal, line 129: Minor nit-pick `s/#/@/`

src/tools/cgcreate.c
```diff
@@ -63,7 +111,11 @@ int main(int argc, char *argv[])
 	uid_t tuid = CGRULE_INVALID, auid = CGRULE_INVALID;
 	gid_t tgid = CGRULE_INVALID, agid = CGRULE_INVALID;
 
+#ifdef WITH_SYSTEMD
+#endif /* WITH_SYSTEMD */
```
kamalesh-babulal, line 115: `ifdef, endif` here aren't guarding any code.

tests/ftests/cgroup.py
drakenclimber: Good find.  I chose to add the mutual exclusion check in libcgroup itself rather than the tests.

src/tools/cgcreate.c
drakenclimber: Oops.  It was a wip - I moved them a few lines above to guard the systemd long option names.

drakenclimber:
v2 changes:
* addressed code review comments
* added `-S` flag to cgcreate to allow the user to set the scope as the default
* add a test for the `-S` option

I'll rebase after the pull request is approved.  It's an easy fix as the only conflict is the ftests Makefile.am

drakenclimber:
Alright, I think this is good to go now.  I added a test for the `domain invalid` logic, and libcgroup seems to be behaving well.  We're relying on the kernel to handle invalid errors, but I wanted to make sure we handled them properly.

kamalesh-babulal:
> Alright, I think this is good to go now. I added a test for the domain invalid logic, and libcgroup seems to be behaving well. We're relying on the kernel to handle invalid errors, but I wanted to make sure we handled them properly.

I reverted the commits:
- api: Add support for empty controller name when building procs path and
- api: Add support for domain invalid cgroups

both the testcases `078` and `079` passed, sorry I guess I am missing something obvious here.

drakenclimber:
> I reverted the commits:
> 
>     * api: Add support for empty controller name when building procs path and
> 
>     * api: Add support for domain invalid cgroups
> 
> 
> both the testcases `078` and `079` passed, sorry I guess I am missing something obvious here.

D'oh.  I pushed the new test to my work-in-progress branch, but I didn't push it to this branch.  My bad.

Commit `api: Add support for empty controller name when building procs path and` is no longer needed and I removed it.

Commit `api: Add support for domain invalid cgroups` is needed for the test I forgot to push - 080-kernel-domain_invalid.py.  But these seem unrelated to this patchset.  Would you like to see them in a separate pull request?

kamalesh-babulal:
> Commit api: Add support for domain invalid cgroups is needed for the test I forgot to push - 080-kernel-domain_invalid.py. But these seem unrelated to this patchset. Would you like to see them in a separate pull request?

I do not have a strong opinion about pushing it with this PR or a new one.

kamalesh-babulal (approved):

drakenclimber:
Pushed to the main and release-3.0 branches.  Thanks so much for the help, @kamalesh-babulal

PR #330, opened 2023-04-10
ftests: cleanup return statements

kamalesh-babulal:
This patch cleans up a few oddities around return statements:
1. Removes redundant return statements across the `ftests/*`, mostly 
    in the function prereqs and a few other places too.
2. Follows the same function template for prereqs, across the `ftests/*.`,
    and removes the checks for success on return from prereqs function,
    those will be true always.
3. Also, replace `consts.TEST_PASSED, None` with `pass` for functions,
    that serves as stub and removes the return value checks for stubs.


drakenclimber (approved):

drakenclimber:
Looks good to me.  Merge to the main and release-3.0 branches.  Thanks, @kamalesh-babulal

PR #331, opened 2023-04-12
Revert enabling controller on cgroup v2 leaf cgroups

kamalesh-babulal:
cgroup v2, has no internal process constraint, where the process runs
only on the leaf node of the cgroup hierarchy and no controllers should
be enabled on the leaf cgroup node too so that they don't compete with
the parent's internal process. This patch broke this rule by enabling
the controller on the leaf node by default. This patch series reverts to
the original approach of not enabling the controller on the leaf cgroup
node and also reverts the logic in the test cases too.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1924,6 +1924,7 @@ error:
 STATIC int cgroupv2_controller_enabled(const char * const cg_name, const char * const ctrl_name)
 {
 	char path[FILENAME_MAX] = {0};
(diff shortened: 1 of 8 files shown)


drakenclimber (approved):

drakenclimber:
Thanks, @kamalesh-babulal for picking this up.  This is the right answer :+1: 

Merged into the main and release-3.0 branches.

PR #332, opened 2023-04-14
ftests: use cgroup Mode for hierarchy check

kamalesh-babulal:
Use `cgroup.get_cgroup_mode()` to check, if the current cgroup
setup mode is as required by the test case to execute.  Some test
cases are very specific to certain cgroup versions. This replaces
the older logic of reading specific controller versions, across the
ftests sources with better Mode check.


drakenclimber:
Nice.  This looks like a really good improvement.  Thanks, @kamalesh-babulal.

What does this mean for hybrid mode?  Now most of the tests will be skipped there.  Do we quit supporting hybrid mode?

kamalesh-babulal:
@drakenclimber Thanks for reviewing, yeah it will effectively skip `Hybrid`.  Check for `HYBRID` that needs to be added, will respin the patch.

kamalesh-babulal:
**V2 Changes:**
- Added check for `HYBRID` mode, while checking for `LEGACY` mode.

kamalesh-babulal:
**v3 Changes:**
- Fix the logic of checking cgroup Modes.

tests/ftests/001-cgget-basic_cgget_v1.py
```diff
@@ -23,9 +23,10 @@ def prereqs(config):
     result = consts.TEST_PASSED
     cause = None
 
-    if CgroupVersion.get_version('cpu') != CgroupVersion.CGROUP_V1:
+    if Cgroup.get_cgroup_mode(config) != (Mode.CGROUP_MODE_LEGACY or
+                                          Mode.CGROUP_MODE_HYBRID):
```
drakenclimber, line 27: I think this is working somewhat by luck.

Doing this `Mode.CGROUP_MODE_LEGACY or Mode.CGROUP_MODE_HYBRID` will always result in an integer `1`.  The `1` is then compared against the result from `get_cgroup_mode()`.  Since LEGACY is `1`, it often succeeds.

This can be verified by comparing the number of hybrid tests skipped in this pull request vs. the number of hybrid tests skipped in the main branch.

This pull request:
https://github.com/libcgroup/libcgroup/actions/runs/4707255489/jobs/8349019318?pr=332

Main branch:
https://github.com/libcgroup/libcgroup/actions/runs/4724261724/jobs/8381266627

PR #334, opened 2023-04-16
tools/cgget: fix '-m' option usage

kamalesh-babulal:
`-m` is logically an independent option and should not be mixed
and matched with other options also when used with `-h, help
always takes precedence over other options. Fix both cases of
usage with a bool flag, that gets set when parsing the command
line option but executes only if it meets the rules of not mixing
with other options.


drakenclimber:

PR #337, opened 2023-04-19
add cgroup v2 support to cgroup_get_current_controller_path()

### LINKED ISSUE #335: cgroup_get_current_controller_path() fails on cgroup v2
On cgroup2, the `/proc/<pid>/cgroup` files look like this:

```sh
$ cat /proc/$$/cgroup
0::/user.slice/user-1000.slice/session-1080.scope
```

Basically, on v2 the "controllers" field is empty, and the `fscanf` in https://github.com/libcgroup/libcgroup/blob/main/src/api.c#L4848 does not accept that as input. The call fails in a rather unfortunate way, returning `ECGOTHER` and "Success", because there is no `errno` set.

kamalesh-babulal:
Add support to read/parse cgroup v2 format `/proc/<pid>/cgroup` to
`cgroup_get_current_controller_path()`. The format of `/proc/<pid>/cgroup`
remains the same:
`hierarchy-ID:controller-list:cgroup-path`

the difference is that in cgroup v2 mode:
- `hierarchy-ID` is always 0 (one hierarchy allowed)
- `controller-list` is empty
    
there can also be a mix of both legacy and unified controllers for a
given task if the cgroup mode is hybrid mode.  This patch, add support
for the mixed cgroup setup mode too.

Closes #335

### DIFF
--- a/src/python/libcgroup.pyx
+++ b/src/python/libcgroup.pyx
@@ -654,6 +654,41 @@ cdef class Cgroup:
         """
         return cgroup.is_cgroup_mode_unified()
 
+    @staticmethod
+    def get_current_controller_path(pid, controller=None):
+        """Get the cgroup path of pid, under controller hierarchy
+
+        Return:
+        Return cgroup path relative to mount point
+
+        Description:
+        Invokes the libcgroup C function, cgroup_get_current_controller_path().
+        It parses the /proc/<pid>/cgroup file and returns the cgroup path, if
+        the controller matches in the output for cgroup v1 controllers and for
+        the cgroup v2 controllers, checks if the cgroup.controllers file has
+        the controller enabled.
+        """
+        cdef char *current_path
+
+        Cgroup.cgroup_init()
+
+        if controller is None:
+            ret = cgroup.cgroup_get_current_controller_path(pid, NULL,
+                        &current_path)
+        elif isinstance(controller, str):
+            ret = cgroup.cgroup_get_current_controller_path(pid,
+                        c_str(controller), &current_path)
+        else:
+            raise TypeError("cgroup_get_current_controller_path failed: "
+                            "expected controller type string, but passed "
+                            "{}".format(type(controller)))
+
+        if ret is not 0:
+            raise RuntimeError("cgroup_get_current_controller_path failed :"
+                               "{}".format(ret))
+
+        return current_path.decode('ascii')
+
     def __dealloc__(self):
         cgroup.cgroup_free(&self._cgp);
 
--- a/tests/ftests/081-pybindings-cgrp_get_curr_ctrl_path-v1.py
+++ b/tests/ftests/081-pybindings-cgrp_get_curr_ctrl_path-v1.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: LGPL-2.1-only
+#
+# cgroup_get_current_controller_path() test using the python bindings (cgroup v1)
+#
+# Copyright (c) 2023 Oracle and/or its affiliates.
+# Author: Kamalesh Babulal <kamalesh.babulal@oracle.com>
+#
+
+from cgroup import Cgroup as CgroupCli, Mode
+from libcgroup import Cgroup, Version
+from process import Process
+import consts
+import ftests
+import sys
+import os
+
+
+CGNAME = '081getctrlpathv1'
+CONTROLLER = 'cpu'
+
+
+def prereqs(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    if config.args.container:
+        result = consts.TEST_SKIPPED
+        cause = 'This test cannot be run within a container'
+        return result, cause
+
+    if Cgroup.cgroup_mode() == Mode.CGROUP_MODE_UNIFIED:
+        result = consts.TEST_SKIPPED
+        cause = 'This test requires the legacy cgroup v1 hierarchy'
+
+    return result, cause
+
+
+def setup(config):
+    CgroupCli.create(config, CONTROLLER, CGNAME)
+
+    config.process.create_process_in_cgroup(config, CONTROLLER, CGNAME, ignore_systemd=True)
+
+
+def test(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    expected_path = "/" + CGNAME
+    pid = CgroupCli.get_pids_in_cgroup(config, CGNAME, CONTROLLER)[0]
+
+    cgrp = Cgroup(CGNAME, Version.CGROUP_V2)
+    #
+    # Test 1 - get the relative path of cgroup, for the pid's cpu controller.
+    #          It's expected to pass because we had created cgroup on cpu
+    #          hierarchy and moved the task to that group.
+    #
+    cgrp_path = cgrp.get_current_controller_path(pid, CONTROLLER)
+    if cgrp_path != expected_path:
+        result = consts.TEST_FAILED
+        cause = 'Expected cgroup path {} got {}'.format(expected_path, cgrp_path)
+
+    #
+    # Test 2 - get the relative path of cgroup, for the pid's memory controller.
+    #          It's expected to fail because we not had created cgroup.
+    #
+    cgrp_path = cgrp.get_current_controller_path(pid, "memory")
+    if cgrp_path == expected_path:
+        result = consts.TEST_FAILED
+        tmp_cause = 'cgroup path unexpectedly formed {}'.format(cgrp_path)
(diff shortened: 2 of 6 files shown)


include/libcgroup/groups.h
```diff
@@ -675,6 +675,12 @@ struct cgroup_controller *cgroup_get_controller_by_index(struct cgroup *cgroup,
  */
 char *cgroup_get_controller_name(struct cgroup_controller *controller);
 
+/**
+ * Return true if cgroup setup mode is cgroup v2 (unified), else
+ * returns false.
+ */
+bool is_unified_mode(void);
```
drakenclimber, line 682: nitpick - we probably should remain consistent with previous APIs and use the `cgroup_` prefix.  How does `cgroup_is_unified_mode()` sound?

drakenclimber (commented):
I think these changes look good.  I tested them out on my v2 box, and they behaved exactly as I hoped/expected.

Let's get some tests around this and add it to the 3.1.0 release.  I especially interested in negative tests and how this performs on the hybrid box.  Thanks, @kamalesh-babulal

include/libcgroup/groups.h
drakenclimber: Silly question - should we just expose `cgroup_setup_mode()` and give the user even more power.  I'm not sure where you're going with this function, so I could be way off base.

kamalesh-babulal: @drakenclimber Thanks for testing and reviewing the patch.  `cgroup_is_unified_mode()` sounds better, as it's consistent with the other APIs and I will add test cases around it.

> Silly question - should we just expose cgroup_setup_mode() and give the user even more power? I'm not sure where you're going with this function, so I could be way off base.

The idea was to export `is_unified_mode()` as a helper to the users, so they can quickly check if the setup is v1 (legacy/hybrid) or v2.  If it returns true it is v2 or v1 (legacy or hybrid) otherwise.  Exposing `cgroup_setup_mode()` could definitely give more power to the user, but that would also mean exposing `enum cg_setup_mode_t`. 

The next question is, why not have `is_legacy_mode()` and `is_hybrid_mode()` functions?

kamalesh-babulal: `enum cg_setup_mode_t` is already exported, will spin up the next version will `cgroup_setup_mode()` and other helpers.

drakenclimber: I don't have a preference either way.  If you think `cgroup_is_unified_mode()` is easier/better for the user, then I'm cool with that.  If we think giving them the full `cg_setup_mode_t` enum is easier, that's fine, too.

kamalesh-babulal:
V2 Changes:
- Removed the `is_mode_unfiied()` commit, its unrelated to this patch.
- Added test cases for both `legacy/hybrid` and `unified` modes.
- Fix the case of `unified` mode and `controller != NULL`

kamalesh-babulal:
**v3 Changes**:
- Rebase the code on top of master branch
- Added the test cases to the `ftests/Makefile.am`
- Added `Reported-by` tag

tests/ftests/081-pybindings-cgrp_get_curr_ctrl_path-v1.py
```diff
@@ -0,0 +1,133 @@
...
+        cause = '\n'.join(filter(None, [cause, tmp_cause]))
+
+    #
+    # Test 3 - get the relative path of cgroup, for the pid's invalid controller.
+    #          It's expected to fail because such controller doesn't exists.
+    #
+    try:
+        cgrp_path = cgrp.get_current_controller_path(pid, "invalid")
+    except RuntimeError as re:
+        if '50011' not in str(re):
```
drakenclimber, line 80: RuntimeError has access to the return code.  You should be able to do this:
```
if re.ret != 50011:
    raise re
```

tests/ftests/081-pybindings-cgrp_get_curr_ctrl_path-v1.py
```diff
@@ -0,0 +1,133 @@
...
+
+    #
+    # Test 4 - get the relative path of cgroup, for the pid's pass NULL as
+    #          controller. It's expected to fail because it's not supported
+    #          cgroup v1.
+    #
+    try:
+        cgrp_path = cgrp.get_current_controller_path(pid, None)
+    except RuntimeError as re:
+        if '50016' not in str(re):
```
drakenclimber, line 91: Same comment as above

tests/ftests/081-pybindings-cgrp_get_curr_ctrl_path-v1.py
```diff
@@ -0,0 +1,133 @@
...
+
+CGNAME = '081getctrlpathv1'
+CONTROLLER = 'cpu'
+
+
+def prereqs(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    if config.args.container:
```
drakenclimber, line 27: Out of curiosity, what happens if this is run in a container?

tests/ftests/081-pybindings-cgrp_get_curr_ctrl_path-v1.py
```diff
@@ -0,0 +1,133 @@
...
+    CgroupCli.create(config, CONTROLLER, CGNAME)
+
+    config.process.create_process_in_cgroup(config, CONTROLLER, CGNAME, ignore_systemd=True)
+
+
+def test(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    expected_path = os.path.join("/" + CGNAME)
```
drakenclimber, line 49: nitpick - I don't think `os.path.join()` is doing anything for us here since we're only providing a single string.

src/python/libcgroup.pyx
```diff
@@ -627,6 +627,40 @@ cdef class Cgroup:
 
         return pid_list
 
+    def get_current_controller_path(self, pid, controller):
```
drakenclimber, line 630: Also, it doesn't appear that we're using the Cgroup instance at all.  Should this method be made a `@staticmethod`?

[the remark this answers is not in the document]
kamalesh-babulal: @drakenclimber Thank you so much for the review. Good idea, it will make cgroup v2 better

kamalesh-babulal: agreed, will make it staticmethod

tests/ftests/081-pybindings-cgrp_get_curr_ctrl_path-v1.py
kamalesh-babulal: `RunError` has access to `ret`, that exception is raised when we run commands, this via pybindings and trying to check for `re.ret` will error out  with `'RuntimeError' object has no attribute 'ret'`

tests/ftests/081-pybindings-cgrp_get_curr_ctrl_path-v1.py
```diff
@@ -0,0 +1,133 @@
...
+    if cgrp_path != expected_path:
+        result = consts.TEST_FAILED
+        cause = 'Expected cgroup path {} got {}'.format(expected_path, cgrp_path)
+
+    #
+    # Test 2 - get the relative path of cgroup, for the pid's memory controller.
+    #          It's expected to fail because we not had created cgroup.
+    #
+    cgrp_path = cgrp.get_current_controller_path(pid, "memory")
+    if cgrp_path == expected_path:
```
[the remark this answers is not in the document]
kamalesh-babulal:
```
# cgget -m
Legacy Mode (Cgroup v1 only).

# cat /proc/1054/cgroup 
13:pids:/user.slice/user-1000.slice/session-1.scope
12:blkio:/user.slice
11:net_cls,net_prio:/
10:memory:/user.slice/user-1000.slice/session-1.scope
9:perf_event:/
8:cpuset:/
7:misc:/
6:rdma:/
5:hugetlb:/
4:devices:/user.slice
3:cpu,cpuacct:/user.slice
2:freezer:/
1:name=systemd:/user.slice/user-1000.slice/session-1.scope
```

systemd might place it in another scope and not necessarily in '/', I too initially assumed it will be `root` cgroup, so limited the check `expected path`

kamalesh-babulal:
**v4 Changes**:
- Addressed the reviews comments from @drakenclimber

tests/ftests/081-pybindings-cgrp_get_curr_ctrl_path-v1.py
drakenclimber: D'oh.  You're totally right.  I obviously confused the two :(

drakenclimber:
Closes #335

drakenclimber:
Pushed to the main branch only.  Thanks @kamalesh-babulal

PR #338, opened 2023-04-23
wrapper: cgroup_add_all_controllers(): fix variable typo

kamalesh-babulal:
Coverity reported uninitialized value:
```
CID 313909 (#1 of 1): Uninitialized scalar variable (UNINIT)8. uninit_use_in_call:
Using uninitialized value *info.name as argument to %s when calling fprintf.
```
In `cgroup_add_all_controllers()`, fix the wrong variable name in the error message,
while parsing the controller name.


drakenclimber (approved):

drakenclimber:

PR #339, opened 2023-04-24
Add cgroup_setup_mode() helpers

kamalesh-babulal:
This patch set exports `cgroup_setup_mode()` and adds following
helpers to check for a cgroup mode, the user might be interested in:
- `is_cgroup_mode_legacy()`
- `is_cgroup_mode_hybrid()`
- `is_cgroup_mode_unified()`

also adds python bindings for the helper functions, along with
test cases to exercise the python bindings.  It also adds a sample
program to demonstrate the usage of `cgroup_setup_mode()`
to the `samples/c/`.

### DIFF
--- a/samples/c/get_setup_mode.c
+++ b/samples/c/get_setup_mode.c
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: LGPL-2.1-only
+/*
+ * Copyright (c) 2023 Oracle and/or its affiliates
+ *
+ * Author: Kamalesh Babulal <kamalesh.babulal@oracle.com>
+ *
+ * Description: This file contains the sample code to demonstrate usage of
+ * 		cgroup_setup_mode() API.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <libcgroup.h>
+
+int main(void)
+{
+	enum cg_setup_mode_t setup_mode;
+	int ret;
+
+	ret = cgroup_init();
+	if (ret) {
+		printf("cgroup_init failed with %s\n", cgroup_strerror(ret));
+		exit(1);
+	}
+
+	setup_mode = cgroup_setup_mode();
+	switch(setup_mode) {
+	case CGROUP_MODE_LEGACY:
+		printf("cgroup mode: Legacy\n");
+		break;
+	case CGROUP_MODE_HYBRID:
+		printf("cgroup mode: Hybrid\n");
+		break;
+	case CGROUP_MODE_UNIFIED:
+		printf("cgroup mode: Unified\n");
+		break;
+	default:
+		printf("cgroup mode: Unknown\n");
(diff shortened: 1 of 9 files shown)


samples/c/get_setup_mode.c
```diff
@@ -0,0 +1,44 @@
...
+/*
+ * Copyright (c) 2023 Oracle and/or its affiliates
+ *
+ * Author: Kamalesh Babulal <kamalesh.babulal@oracle.com>
+ *
+ * Description: This file contains the sample code to demonstrate usage of
+ * 		cgroup_setup_mode() API.
+ */
+
+#include "../src/libcgroup-internal.h"
```
drakenclimber, line 11: I wouldn't expect an external program to need to include `libcgroup-internal.h`.  Shouldn't `#include <libcgroup.h>` be sufficient?

samples/c/get_setup_mode.c
```diff
@@ -0,0 +1,44 @@
...
+
+	ret = cgroup_init();
+	if (ret) {
+		printf("cgroup_init failed with %s\n", cgroup_strerror(ret));
+		exit(1);
+	}
+
+	setup_mode = cgroup_setup_mode();
+	switch(setup_mode) {
+	case 1:
```
drakenclimber, line 29: `CGROUP_MODE_LEGACY` is available via the public header `libcgroup.h`.  Shouldn't a user use it rather than the magic number `1`?

tests/ftests/083-pybindings-helpers_cgroup_mode.py
```diff
@@ -0,0 +1,79 @@
...
+
+    mode1 = Cgroup.cgroup_mode()
+    mode2 = CgroupCli.get_cgroup_mode(config)
+
+    if mode1 != mode2:
+        result = consts.TEST_FAILED
+        cause = 'mode mismatch: libcgroup mode: {}, tests mode: {}'.format(mode1, mode2)
+        return result, cause
+
+    if mode1 == 1:
```
drakenclimber, line 38: The modes are available in Python via `Mode.CGROUP_MODE_LEGACY`, etc.; we should use them rather than magic numbers.  (You'll need to update the import to be `from libcgroup import Cgroup, Mode`)

samples/c/get_setup_mode.c
kamalesh-babulal: @drakenclimber Thank you so much for the review.  Will re-spin v2 addressing all comments.

drakenclimber (approved):

drakenclimber:
Pushed to the main branch.  I _didn't_ push it to the release-3.0 branch since there were merge conflicts.  Release 3.1 will be based upon the main branch, so we don't need the change in the release-3.0 branch.

Thanks @kamalesh-babulal

PR #340, opened 2023-04-26
cgcreate: add -p <pid> support for systemd scope 

kamalesh-babulal:
Currently, an idle_thread is created for the transient systemd scope by
default, it works fine, when the slice/scope creation is treated in par
to create of traditional cgroup, but the user may wish to create a
slice/scope, with a process of their choice, instead of a default idle
thread.  This patchset adds support to pass task pid via `-p`  option,
while creating a systemd slice/scope.  Also, adds support for ftest
cgroup to accept pid during scope creation and the ftest test case
to exercise it.

### DIFF
--- a/doc/man/cgcreate.1
+++ b/doc/man/cgcreate.1
@@ -33,7 +33,8 @@ creates a new systemd scope. The cgroup name provided after the
 \fB-g\fR flag must be of the form
 \fB<slice-name>.slice/<scope-name>.scope\fR. If the slice
 does not exist, systemd will create it. Libcgroup will place an
-idle process in the scope's cgroup.procs file.
+idle process in the scope's cgroup.procs file unless the \fB-p\fR
+flag is provided.
 
 .TP
 .B -d, --dperm=mode
@@ -62,6 +63,12 @@ multiple times.
 .TP
 .B -h, --help
 display this help and exit
+
+.TP
+.B -p, --pid=pid
+moves the provided \fBpid\fR into the
+\fB<slice-name>.slice/<scope-name>.scope\fR.  Must be used in
+conjunction with \fB-c\fR.
 
 .TP
 .B -s, --tperm=mode
--- a/src/tools/cgcreate.c
+++ b/src/tools/cgcreate.c
@@ -45,6 +45,8 @@ static void usage(int status, const char *program_name)
 	info("  -b				Ignore default systemd");
 	info("delegate hierarchy\n");
 	info("  -c, --scope			Create a delegated systemd scope\n");
+	info("  -p, --pid=pid                   Task pid to use to create systemd ");
+	info("scope\n");
 	info("  -S, --setdefault		Set this scope as the default scope ");
 	info("delegate hierarchy\n");
 #endif
@@ -52,7 +54,7 @@ static void usage(int status, const char *program_name)
 
 #ifdef WITH_SYSTEMD
 static int create_systemd_scope(struct cgroup * const cg, const char * const prog_name,
-				int set_default)
+				int set_default, pid_t pid)
 {
 	struct cgroup_systemd_scope_opts opts;
 	char slice[FILENAME_MAX];
@@ -63,6 +65,8 @@ static int create_systemd_scope(struct cgroup * const cg, const char * const pro
 	ret = cgroup_set_default_scope_opts(&opts);
 	if (ret)
 		return ret;
+
+	opts.pid = pid;
 
 	ret = cgroup_create_scope2(cg, 0, &opts);
 	if (!ret && set_default) {
@@ -115,6 +119,7 @@ int main(int argc, char *argv[])
 #ifdef WITH_SYSTEMD
 		{"scope",	      no_argument, NULL, 'c'},
 		{"setdefault",	      no_argument, NULL, 'S'},
+		{"pid",		required_argument, NULL, 'p'},
 #endif /* WITH_SYSTEMD */
 		{0, 0, 0, 0},
 	};
@@ -123,8 +128,9 @@ int main(int argc, char *argv[])
 	gid_t tgid = CGRULE_INVALID, agid = CGRULE_INVALID;
 
 	int ignore_default_systemd_delegate_slice = 0;
-	int create_scope = 0;
 	int set_default_scope = 0;
+	int create_scope = 0;
+	pid_t scope_pid = -1;
 
 	struct cgroup_group_spec **cgroup_list;
 	struct cgroup_controller *cgc;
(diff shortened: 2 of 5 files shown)


tests/ftests/084-sudo-cgcreate_systemd_scope_pid.py
```diff
@@ -0,0 +1,148 @@
...
+        pid = int(Cgroup.get(config, None, CGNAME1, setting='cgroup.procs',
+                             print_headers=False, values_only=True, ignore_systemd=True))
+        # use the pid variable so that lint is happy
+        Log.log_debug('Cgroup {} has pid {}'.format(CGNAME1, pid))
+    except RunError:
+        result = consts.TEST_FAILED
+        cause = 'Failed to read pid in {}\'s cgroup.procs'.format(CGNAME1)
+        return result, cause
+
+    if scope_pid != pid:
```
drakenclimber, line 79: I'm not sure how Python isn't yelling at us here.  `pid` is declared inside of the `try` statement above, but is being tested outside of it.  If you move this `if` inside of the `try` statement, I think you can remove the `log_debug` message above.

tests/ftests/084-sudo-cgcreate_systemd_scope_pid.py
```diff
@@ -0,0 +1,148 @@
...
+
+    #
+    # Test 3: Pass the already created task pid as scope_pid for a new systemd
+    #         scope.  This would mean the CGNAME1 should be killed and the pid
+    #         should be the default scope task of CGNAME2
+    Cgroup.create_and_validate(config, CONTROLLERS, CGNAME2, create_scope=True, scope_pid=scope_pid)
+
+    # CGNAME1 should be deleted by the systemd.
+    try:
+        pid = int(Cgroup.get(config, None, CGNAME1, setting='cgroup.procs',
```
drakenclimber, line 92: Same comment as above - use the helper function.

tests/ftests/084-sudo-cgcreate_systemd_scope_pid.py
```diff
@@ -0,0 +1,148 @@
...
+                             print_headers=False, values_only=True, ignore_systemd=True))
+    except RunError as re:
```
drakenclimber, line 94: You'll probably need an `else` statement here in cause the reading of `cgroup.procs` erroneously succeeds.

```
try:
    ...
except RunError as re:
    ...
else:
    result = consts.TEST_FAILED
    cause = 'Erroneously succeeded reading cgroup.procs in {}'.format(CGNAME1)
    return result, cause
```

tests/ftests/084-sudo-cgcreate_systemd_scope_pid.py
```diff
@@ -0,0 +1,148 @@
...
+        pid = int(Cgroup.get(config, None, CGNAME2, setting='cgroup.procs',
+                             print_headers=False, values_only=True, ignore_systemd=True))
+        # use the pid variable so that lint is happy
+        Log.log_debug('Cgroup {} has pid {}'.format(CGNAME2, pid))
+    except RunError:
+        result = consts.TEST_FAILED
+        cause = 'Failed to read pid in {}\'s cgroup.procs'.format(CGNAME2)
+        return result, cause
+
+    if scope_pid != pid:
```
drakenclimber, line 108: Same comment as above.  Can we move this `if` inside the `try` and delete the `log_debug()` message?

tests/ftests/084-sudo-cgcreate_systemd_scope_pid.py
```diff
@@ -0,0 +1,148 @@
...
+        result = consts.TEST_FAILED
+        cause = 'scope created with other pid {}, expected pid {}'.format(pid, scope_pid)
+
+    return result, cause
+
+
+def teardown(config):
+    pid = int(Cgroup.get(config, None, CGNAME2, setting='cgroup.procs',
```
drakenclimber, line 116: Ditto - use helper function

doc/man/cgcreate.1
```diff
@@ -63,6 +63,11 @@ multiple times.
 .B -h, --help
 display this help and exit
 
+.TP
+.B -p <pid>
+uses the pid specified as the default process, in place of the idle
```
drakenclimber, line 68: Thinking out loud here: I would be tempted to not mention the idle process at all.  Perhaps something like this:

```
moves the specified \fB<pid>\fR into the
\fB<slice-name>.slice/<scope-name>.scope\fR.  Must be used in
conjunction with \fB-c\fR.
````

We should update the `-c` flag as well.  Perhaps something like:

```
.B -c, --scope
creates a new systemd scope. The cgroup name provided after the
\fB-g\fR flag must be of the form
\fB<slice-name>.slice/<scope-name>.scope\fR. If the slice
does not exist, systemd will create it. Libcgroup will place an
idle process in the scope's cgroup.procs file unless the \fB-p\fR
flag is provided
```

File: src/tools/cgcreate.c [hunk 7]
```diff
@@ -45,14 +45,16 @@ static void usage(int status, const char *program_name)
 	info("  -b				Ignore default systemd");
 	info("delegate hierarchy\n");
 	info("  -c, --scope			Create a delegated systemd scope\n");
+	info("  -p <pid>			Task pid to use to create systemd ");
```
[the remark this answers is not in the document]
kamalesh-babulal: @drakenclimber Thank you so much for the review.  We can have the long option too.

tests/ftests/084-sudo-cgcreate_systemd_scope_pid.py
```diff
@@ -0,0 +1,148 @@
...
+    #
+    # Test 2: Pass a valid task pid as scope_pid and systemd scope creation
+    #         will succeed. Read the scope cgroup.procs to find if the task
+    #         passed by us was used as scope task.
+    #
+    scope_pid = config.process.create_process(config)
+    Cgroup.create_and_validate(config, CONTROLLERS, CGNAME1, create_scope=True, scope_pid=scope_pid)
+
+    try:
+        pid = int(Cgroup.get(config, None, CGNAME1, setting='cgroup.procs',
```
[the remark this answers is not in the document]
kamalesh-babulal: The helper is better, will replace it at all call sites.

doc/man/cgcreate.1
kamalesh-babulal: Makes sense to update `-c` option, your rephrase looks better, I will make the man page change accordingly.

tests/ftests/084-sudo-cgcreate_systemd_scope_pid.py
kamalesh-babulal: Sure, will move the code.

tests/ftests/084-sudo-cgcreate_systemd_scope_pid.py
kamalesh-babulal: missed the erroneously read case, will add it

kamalesh-babulal:
**V2 Changes:**
- rebased on top of `main` branch
- Addressed review comments from @drakenclimber

drakenclimber (approved):

drakenclimber:
Pushed to the main branch.  Thanks for the patchset, @kamalesh-babulal

PR #341, opened 2023-05-02
[RFC] tools/cgexec: add -r option for systemd scope

kamalesh-babulal:
This patch set provides an option for users to replace the default `idle_thread`
created for systemd scope.  It introduces '-r' option, when specified, the program
passed as the argument to `cgexec` will replace the default `idle_thread`, if it
exists.  Also, if it's the only task, killing it will remove the .scope cgroup it's running
in.  It also adds support for ftests to pass the `-r` flag and a test case to exercise the
different likely and unlikely paths with replace flag.


tests/ftests/cgroup.py
```diff
@@ -978,6 +978,9 @@ def cgexec(config, controller, cgname, cmdline, sticky=False,
         if (ignore_systemd):
             cmd.append('-b')
 
+        if(replace_idle):
```
kamalesh-babulal, line 981: Will fix the `lint` warning in the next revision.

src/tools/cgexec.c
```diff
@@ -51,28 +55,41 @@ static void usage(int status, const char *program_name)
 	info("change pidlist and children tasks\n");
 #ifdef WITH_SYSTEMD
 	info("  -b				Ignore default systemd delegate hierarchy\n");
+	info("  -r				Replace the systemd scope idle process\n");
```
drakenclimber, line 58: Hmmm.... We should probably be consistent in what we call the idle thread.  (It looks like in the cgcreate man page we call it an "idle process".  FWIW, the actual executable is named `libcgroup_systemd_idle_thread`)

src/tools/cgexec.c
```diff
@@ -171,12 +188,128 @@ int main(int argc, char *argv[])
...
 	}
 
-	/* Now exec the new process */
-	ret = execvp(argv[optind], &argv[optind]);
-	if (ret == -1) {
-		err("%s", strerror(errno));
+	if (!replace_idle) {
+		/* Now exec the new process */
+		ret = execvp(argv[optind], &argv[optind]);
+		err("exec failed:%s", strerror(errno));
```
drakenclimber, line 194: nitpick - looks like the `if (ret == -1)` was removed, and now this error message will always be printed.

src/tools/cgexec.c
```diff
@@ -171,12 +188,128 @@ int main(int argc, char *argv[])
...
+	exec_task_pid = fork();
+	if (exec_task_pid == -1) {
+		err("Fork failed:%s\n", strerror(errno));
+		return  -1;
+	}
+
+	/* child process exec's into task program */
+	if (exec_task_pid == 0) {
+		execvp(argv[optind], &argv[optind]);
+		err("exec failed:%s", strerror(errno));
```
drakenclimber, line 218: Same comment as above.  Do we want to surround this with an `if (ret == -1)`

src/tools/cgexec.c
```diff
@@ -171,12 +188,128 @@ int main(int argc, char *argv[])
...
+			ret = sscanf(buffer, "%d:%[^:]:%4096s\n", &id, ctrl_name, cgroup_name);
+
+		/* skip named controllers */
+		if (ret == 3 && strncmp(ctrl_name, "name", 4) == 0)
+			continue;
+
+		/* skip if the cgroup doesn't ends with .scope */
+		base = basename(cgroup_name);
+		len = strlen(base) - 6;
+		if (strcmp(base + len, ".scope") != 0)
```
drakenclimber, line 247: Do we need to worry about buffer underflows or overflows here?

I can't see how a buffer overflow can happen, but it seems like an underflow could happen.  What if the `cgroup_name` is say `foo` and thus less than 6 characters long.  In such a case, `len` would become negative, correct?

src/tools/cgexec.c
```diff
@@ -171,12 +188,128 @@ int main(int argc, char *argv[])
...
+		/* cgroup v2 (unified) mode */
+		if (ret == 2) {
+			scope_pid = find_scope_idle_task_pid(cgroup_name, NULL);
+		} else {
+			/* cgroup v1 (legacy/hybrid) mode */
+			/* check if the mount point is shared between controllers */
+			ctrl = strchr(ctrl_name, ',');
+			if (ctrl) {
+				/* truncate other controllers, but first */
+				len = strlen(ctrl_name) - strlen(ctrl);
```
drakenclimber, line 259: This looks like it works, but wouldn't it be simpler to just do:
`ctrl[0] = '\0';`

src/tools/cgexec.c
```diff
@@ -171,12 +188,128 @@ int main(int argc, char *argv[])
...
+	for(i = 0; i < size; i++) {
+		snprintf(pid_proc_cmd_path, FILENAME_MAX, "/proc/%u/cmdline", pids[i]);
+		pid_proc_cmd_fp = fopen(pid_proc_cmd_path, "re");
+		/* task might have exited */
+		if (!pid_proc_cmd_fp)
+			continue;
+
+		/* task might have exited, so consider only successful reads. */
+		if (fgets(buffer, FILENAME_MAX, pid_proc_cmd_fp)) {
+			if (strcmp(buffer, "libcgroup_systemd_idle_thread") == 0) {
```
drakenclimber, line 307: Should we make `libcgroup_systemd_idle_thread` a constant?

src/tools/cgexec.c
```diff
@@ -171,12 +188,128 @@ int main(int argc, char *argv[])
...
+				return  -1;
+			}
+		}
+	}
+	fclose(pid_proc_cgrp_fp);
+
 	return 0;
 }
+
+static int find_scope_idle_task_pid(char *cgroup_name, char *ctrl_name)
```
drakenclimber, line 285: nitpick - should this return a `pid_t`?

doc/man/cgexec.1
```diff
@@ -49,6 +49,11 @@ kernel values.
 .B -h, --help
 Display this help and exit.
 
+.TP
+.B -r
+Replaces systemd scope's idle process with the task \fBcommand\fR.
```
drakenclimber, line 54: Same comment as earlier.  We need to use consistent nomenclature for the idle thread.

Now what that should be?... Well, I'm not sure yet :).

src/tools/cgexec.c
kamalesh-babulal: Agreed, will fix it in the next version.

src/tools/cgexec.c
kamalesh-babulal: agreed and will fix it.

src/tools/cgexec.c
kamalesh-babulal: It's trying to extract the first controller. for example, `cpu,cpuacct`, truncate `,cpuacct`

src/tools/cgexec.c
kamalesh-babulal: Agreed, we need it consistent term. Will try and come up with one.

src/tools/cgexec.c
kamalesh-babulal: Good idea, sure.

src/tools/cgexec.c
kamalesh-babulal: Does `libcgroup_systemd_idle_process` make sense? `thread` vs `process` might be confusing for some.  Let's use `idle process` ?

kamalesh-babulal:
**V2 Changes**
- Rewrite the logic to support the `v1 (legacy/hybrid)` and `v2`
- Addressed some of the review comments

PR #349, opened 2023-05-18
Add an example that utilizes the systemd C APIs

drakenclimber:


src/api.c
```diff
@@ -4619,9 +4620,24 @@ int cgroup_change_cgroup_path(const char *dest, pid_t pid, const char *const con
 	}
 	memset(&cgroup, 0, sizeof(struct cgroup));
 
-	ret = cg_prepare_cgroup(&cgroup, pid, dest, controllers);
-	if (ret)
-		return ret;
+	mode = cgroup_setup_mode();
+
+	if (mode == CGROUP_MODE_UNIFIED && !controllers) {
```
kamalesh-babulal, line 4625: `is_cgroup_mode_unified()` can be used here.

src/python/libcgroup.pyx
```diff
@@ -689,6 +689,27 @@ cdef class Cgroup:
 
         return current_path.decode('ascii')
 
+    @staticmethod
+    def move_proc(pid, dest_cgname):
```
kamalesh-babulal, line 693: Do you think using `process` or `task` instead of `proc` make it more intuitive?

.github/workflows/continuous-integration.yml
```diff
@@ -142,6 +142,24 @@ jobs:
...
+    - uses: actions/checkout@v3
+      with:
+        submodules: false
+    - name: Initialize the directory
+      uses: ./.github/actions/setup-libcgroup
+    - name: Reconfigure libcgroup with samples enabled
+      run: CFLAGS="$CFLAGS -Werror" ./configure --sysconfdir=/etc --localstatedir=/var --enable-code-coverage --enable-opaque-hierarchy="name=systemd" --enable-python --enable-samples
+    - name: Build the library
+      run: make
+
```
kamalesh-babulal, line 162: Should we have a step `buildtest` combining both `buildsamples` and `withoutsystemd`, to avoid spawning a new VM for build tests?

samples/c/create_systemd_scope.c
```diff
@@ -0,0 +1,614 @@
...
+			opts->debug = true;
+			break;
+		default:
+			usage(argv[0]);
+			exit(1);
+			break;
+		}
+	}
+
+	return ret;
```
kamalesh-babulal, line 117: Minor nitpick `ret` is unused, should it be `return 0`?

samples/c/create_systemd_scope.c
```diff
@@ -0,0 +1,614 @@
...
+	} else if (*pid == 0) {
+		/*
+		 * This is the child process.  Move it to the requested cgroup
+		 */
+		ret = cgroup_change_cgroup_path(cgname, getpid(), NULL);
+		if (ret)
+			exit(0);
+
+		while (true)
+			sleep(1);
```
kamalesh-babulal, line 453: Should the sleep wake-up seconds be increased to larger number, instead of waking up every second

samples/c/create_systemd_scope.c
```diff
@@ -0,0 +1,614 @@
...
+	       "scope.  Systemd will not automatically delete the slice.\n");
+	printf("----------------------------------------------------------------\n\n");
+
+	wait_for_child(high_pid);
+	wait_for_child(med_pid);
+	wait_for_child(low_pid);
+
+error:
+	return ret;
+}
```
kamalesh-babulal, line 614: @drakenclimber This is a pretty cool training manual, instead of a sample. Thank you so much for making it so extensive.

src/python/libcgroup.pyx
drakenclimber: Honestly I'm not sure.  We already have a python method, `get_procs()`, so I was trying to be consistent with that name.  What if I updated both:
* `move_procs()` -> `move_processes()`
* `get_procs()` -> `get_processes()`

The python multiprocessing library seems to use `process` fairly heavily
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.current_process

.github/workflows/continuous-integration.yml
drakenclimber: That's a good plan :+1:

samples/c/create_systemd_scope.c
drakenclimber: Yeah, that makes sense.  I'll bump it up to `10` seconds.

samples/c/create_systemd_scope.c
drakenclimber: Good call.  I'll change it.

kamalesh-babulal (approved):

src/python/libcgroup.pyx
kamalesh-babulal: `move_process()` and `get_processes()` make sense. Feel free to merge the patchset with the minor change.

drakenclimber:
Pushed to main.  Thanks for the review, @kamalesh-babulal.

One minor note - this sample program requires `WITH_SYSTEMD` to be defined, so I wasn't able to combine the `--enable-samples` github job with the `--enable-systemd=no` job. :(

kamalesh-babulal:
@drakenclimber How about compiling that one program when `systemd` is enabled:
https://github.com/libcgroup/libcgroup/commit/5e6d7214d060cf08fa5628273a9fa6de71845d7f
https://github.com/libcgroup/libcgroup/commit/3224b68d0e1b5ac8fef97ee71002b38dc03498de

PR #350, opened 2023-05-22
tests-999: add a random directory to mount path

kamalesh-babulal:
Currently, the mount path is static/hard coded and sometimes there might be corner cases where the mount path directories are not deleted completely, or there might be two instances of CI running at the same time. Let's introduce an additional directory to the mount path, that's generated randomly so that every run is unique and doesn't interfere with each other.

```
----------------------------------------------------------------- 
Test Results:
        Run Date:                          May 22 11:27:45
        Passed:                                  1 test(s)
        Skipped:                                 0 test(s)
        Failed:                                  0 test(s)
-----------------------------------------------------------------
Timing Results:
        Test                          Time (sec)
        ----------------------------------------
        setup                               0.00
        999-stress-cgroup_init.py           1.36
        teardown                            0.00
        ----------------------------------------
        Total Run Time                      1.36
```

### DIFF
--- a/tests/ftests/999-stress-cgroup_init.py
+++ b/tests/ftests/999-stress-cgroup_init.py
@@ -11,16 +11,19 @@ from libcgroup import Cgroup
 from run import Run
 import consts
 import ftests
+import random
+import string
 import sys
 import os
 
 MNT_COUNT = 101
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:

PR #352, opened 2023-05-25
samples/Makefile: introduce systemd dependency

kamalesh-babulal:
Some of the `samples` programs depend on `systemd` to be
enabled during configuration.  Add this dependency to the
`Makefile` so that the samples are built according to the current
configuration options.  This patch also, folds different CI workflow
build tests into single step and adds `systemd=no, samples=yes`
case too.


drakenclimber (approved):

drakenclimber:

PR #353, opened 2023-05-25
Add systemd python example

drakenclimber:


src/python/libcgroup.pyx
```diff
@@ -37,6 +37,12 @@ cdef class SystemdMode:
     CGROUP_SYSTEMD_MODE_IGNORE_DEPS = cgroup.CGROUP_SYSTEMD_MODE_IGNORE_DEPS
     CGROUP_SYSTEMD_MODE_IGNORE_REQS = cgroup.CGROUP_SYSTEMD_MODE_IGNORE_REQS
 
+cdef class LogLevel:
+    CGROUP_LOG_ERROR = cgroup.CGROUP_LOG_ERROR
```
kamalesh-babulal, line 45: Minor nitpick, the `CGROUP_LOG_CONT` is not be used.

samples/python/create_systemd_scope.py
```diff
@@ -0,0 +1,185 @@
...
+#
+
+#
+# To run this program:
+#       1. Compile the libcgroup library
+#           $ ./bootstrap.sh
+#           $ make
+#       2. Note the path to your libcgroup python library
+#           $ pushd src/python/build/lib*
+#           $ pwd
```
kamalesh-babulal, line 23: Should this be `export PYTHONPATH=$(pwd)`

tests/ftests/071-sudo-set_default_systemd_cgroup.py
```diff
@@ -42,131 +36,144 @@ def prereqs(config):
...
+    #
+    # Test 2 - write_default_systemd_scope() should succeed if the slice/scope
+    #          are invalid and we're not setting it as the default.  If we
+    #          create a cgroup at this point, it should be created at the root
+    #          cgroup level, and the default slice/scope should have no bearing
+    #
+    Cgroup.write_default_systemd_scope(SLICE, SCOPE, False)
+
+    pid = config.process.create_process(config)
+    cg = Cgroup(OTHER_CGNAME, Version.CGROUP_V2)
```
kamalesh-babulal, line 71: Should the check for the `cgroup setup mode()` for `cgroup v2` be added in the `prereqs`, as the test case explicitly uses cgroup v2

tests/ftests/071-sudo-set_default_systemd_cgroup.py
```diff
@@ -42,131 +36,144 @@ def prereqs(config):
...
+        result = consts.TEST_FAILED
+        tmp_cause = 'write_default_systemd_scope() erroneously passed'
+        cause = '\n'.join(filter(None, [cause, tmp_cause]))
 
+    #
+    # Test 4 - Create a systemd scope and set it as the default.  Everything
+    #          should work properly in this case
+    #
+    pid = None
+    if Cgroup.cgroup_mode() != Mode.CGROUP_MODE_LEGACY:
```
kamalesh-babulal, line 117: This will include `hybrid` and `unified`, it can be rewritten to `if Cgroup.cgroup_mode() == Mode.CGROUP_MODE_UNIFIED` to check only for `unified`

tests/ftests/071-sudo-set_default_systemd_cgroup.py
```diff
@@ -42,131 +36,144 @@ def prereqs(config):
...
+        pid = config.process.create_process(config)
+        Cgroup.create_scope(SCOPE, SLICE, pid=pid)
```
kamalesh-babulal, line 119: The initial idea was to emulate `create_scope()`,  so that users can use one of their hand-created `slice` and `scope`

src/python/libcgroup.pyx
```diff
@@ -615,6 +615,25 @@ cdef class Cgroup:
...
+        Also delete the internal global variable within libcgroup so that
+        the default path is restored back to the root cgroup
+        """
+        try:
+            Cgroup.write_default_systemd_scope('', '')
+        except RuntimeError:
+            pass
+
+        try:
+            Cgroup.__set_default_systemd_cgroup()
```
kamalesh-babulal, line 637: I guess `__set_default_systemd_cgroup()` is redundant here. The above `write_default_systemd_scope`, will call `__set_default_systemd_cgroup()` internally by default

samples/python/create_systemd_scope.py
```diff
@@ -0,0 +1,185 @@
...
+          '\t  but it has been removed by this program')
+    print('\t* PID {} has been placed in the {} cgroup'.format(high_pid, HIGH_CGNAME))
+    print('\t* PID {} has been placed in the {} cgroup'.format(med_pid, MED_CGNAME))
+    print('\t* PID {} has been placed in the {} cgroup'.format(low_pid, LOW_CGNAME))
+    print('\nThis program will wait for the aforementioned child processes to\n'
+          'exit before exiting itself. Systemd will automatically delete the\n'
+          'scope when there are no longer any processes running within the\n'
+          'scope.  Systemd will not automatically delete the slice.')
+    print('----------------------------------------------------------------\n')
+
```
kamalesh-babulal, line 179: Should we explicitly kill the `{high, med, low}_pid` ?

tests/ftests/071-sudo-set_default_systemd_cgroup.py
drakenclimber: I don't think we need to limit this test to cgroup v2 only.  When a Cgroup class is instantiated via `Cgroup('TomsCgroup', Version.CGROUP_V2)`, the user is telling libcgroup that they will be speaking v2 format.  The library can/will convert the cgroup to v1 or hybrid if the user requests as such.

And my latest test run shows that this successfully runs on v1, hybrid, and v2:
https://github.com/drakenclimber/libcgroup/actions/runs/5271801648/jobs/9533192309

tests/ftests/071-sudo-set_default_systemd_cgroup.py
drakenclimber: I wrote this a couple weeks ago, and I don't remember the exact details/reasons, but the code within the `if` fails on legacy but works on hybrid and unified.

tests/ftests/071-sudo-set_default_systemd_cgroup.py
drakenclimber: I see.  Do you want me to rework this to match your original goal?  Do these changes sufficiently test the code?

src/python/libcgroup.pyx
drakenclimber: On a successful case, `__set_default_systemd_cgroup()` will get definitely get called by `write_default_systemd_scope()`, but I was afraid of the case where the call to `cgroup_write_systemd_default_cgroup()` failed and thus an exception was raised before the default was overwritten.  Is this crazy?

samples/python/create_systemd_scope.py
kamalesh-babulal: Make sense, so that they can refer to it.

src/python/libcgroup.pyx
kamalesh-babulal: Please consider this:
`write_default_systemd_scope()` `(cgroup_write_systemd_default_cgroup)` can fail at:
1. Fail to create `/var/run/libcgroup` directory or
2. Fail to open `/var/run/libcgroup/systemd` file for write or
3. Fail to write `slice/scope` into `/var/run/libcgroup/systemd`

In the above cases, the `default_systemd_scope()` if set before calling `clear_default_systemd_scope()` will remain set, and calling `Cgroup.__set_default_systemd_cgroup()` might set it again. Did I miss something here?

tests/ftests/071-sudo-set_default_systemd_cgroup.py
kamalesh-babulal: True, sorry for the noise.

tests/ftests/071-sudo-set_default_systemd_cgroup.py
kamalesh-babulal: oh okay, yeah it will work on the `hybrid/unified`

tests/ftests/071-sudo-set_default_systemd_cgroup.py
kamalesh-babulal: The changes will test the `default systemd slice` so I don't have a strong opinion. We should be good, with `create_scope()`

src/python/libcgroup.pyx
drakenclimber: Ooohhh... good counterpoint.  What if we do this?  `__set_default_systemd_cgroup()` will only get called once, but we guarantee it's always called.

```
        try:
            Cgroup.write_default_systemd_scope('', '', set_default=False) 
        except RuntimeError:
            pass

        try:
            Cgroup.__set_default_systemd_cgroup()
        except RuntimeError:
            pass

```

kamalesh-babulal: This change looks good, the redundancy goes away.

kamalesh-babulal (approved):

drakenclimber:

PR #354, opened 2023-05-26
[RFC] tools/cgexec: add -r option for systemd scope

kamalesh-babulal:
This patch set provides an option for users to replace the default idle_thread
created for systemd scope. It introduces '-r' option, when specified, the program
passed as the argument to cgexec will replace the default idle_thread, if it
exists. Also, if it's the only task, killing it will remove the .scope cgroup it's running
in. It also adds support for ftests to pass the -r flag and a test case to exercise the
different likely and unlikely paths with replace flag.

PR #355, opened 2023-06-02
[RFC] tools/cgclassify: add -r option for systemd scope

kamalesh-babulal:
This patch set provides an option for users to replace the default idle_thread
created for systemd scope. It introduces '-r' option, when specified, the first
pid passed as the argument to cgclassify will replace the default idle_thread,
if it exists.  Also, if it's the only task, killing it will remove the .scope cgroup it's
running in. It also adds support for ftests to pass the -r flag and a test case to
exercise the different likely and unlikely paths with replace flag.

### DIFF
--- a/tests/ftests/999-stress-cgroup_init.py
+++ b/tests/ftests/999-stress-cgroup_init.py
@@ -17,9 +17,9 @@ import sys
 import os
 
 MNT_COUNT = 101
-MNT_POINT = '/tmp/999stress/'
+MNT_POINT = '/tmp/'
 DIR_PREFIX = 'name'
-RANDOM_DIR = ''.join(random.choices(string.ascii_letters, k=5))
+RANDOM_DIR = ''.join(random.choices(string.ascii_letters, k=5)) + '-999stress'
 
 
 def cgroup_path(count):
@@ -33,7 +33,6 @@ def prereqs(config):
 def setup(config):
     cmd = ['sudo', 'mkdir']
 
-    cmd.append(MNT_POINT)
     cmd.append(MNT_POINT + RANDOM_DIR)
 
     for count in range(MNT_COUNT):
@@ -78,7 +77,6 @@ def teardown(config):
         cmd.append(cgroup_path(count))
 
     cmd.append(MNT_POINT + RANDOM_DIR)
-    cmd.append(MNT_POINT)
 
     # execute rmdir top-level top-level/sub-directory* at once.
     Run.run(cmd)


kamalesh-babulal:
The first patches are from #354, this patchset is based on parts of it. I will remove them when `cgexec` patches get merged.

src/tools/cgclassify.c
```diff
@@ -198,7 +231,449 @@ int main(int argc, char *argv[])
...
+		if (_ctrl_name) {
+			size = strlen(ctrl_name) - strlen(_ctrl_name);
+			ctrl_name[size] = '\0';
+		}
+
+		/*
+		 * capture is true, while the pid's controller and cgroups
+		 * are populated for rollback case.
+		 */
+		if (capture) {
```
drakenclimber, line 345: We should probably add a sanity check here:

```
if (i > MAX_MNT_ELEMENTS) {
    err("something or other\n");
    // I'm not sure if we should set the scope_pid back to -1 or do other things
    goto out;
}
```

tests/ftests/088-sudo-cgclassify_systemd_scope.py
```diff
@@ -0,0 +1,169 @@
...
+from run import RunError
+import consts
+import ftests
+import time
+import sys
+import os
+
+CONTROLLER = 'cpu'
+SLICE = 'libcgroup.slice'
+CGNAME = os.path.join(SLICE, '086cgcreate.scope')
```
drakenclimber, line 21: nitpick - test number 086 isn't correct :)

tests/ftests/088-sudo-cgclassify_systemd_scope.py
```diff
@@ -0,0 +1,169 @@
...
+    result = consts.TEST_PASSED
+    cause = None
+
+    #
+    # Test 1: call cgclassify -r to replace idle_thread with infinite while loop
+    #
+    Cgroup.create_and_validate(config, CONTROLLER, CGNAME, create_scope=True)
+    idle_pid = Cgroup.get_pids_in_cgroup(config, CGNAME, CONTROLLER)[0]
+
+    config.process.create_process_in_cgroup(config, CONTROLLER, CGNAME, replace_idle=True)
```
drakenclimber, line 51: Since this test is designed to test cgclassify, you may want to explicitly specify that (in case the default ever changes):
```
config.process.create_process_in_cgroup(config, CONTROLLER, CGNAME, cgclassify=True, replace_idle=True)
```

tests/ftests/088-sudo-cgclassify_systemd_scope.py
```diff
@@ -0,0 +1,169 @@
...
+
+    #
+    # Test 2: call cgclassify -r to replace non-idle_thread (while loop,
+    #         previously created) with another while loop.
+    #
+    try:
+        config.process.create_process_in_cgroup(config, CONTROLLER, CGNAME, replace_idle=True)
+    except RunError as re:
+        if "Failed to find idle_thread task" not in re.stderr:
+            raise re
```
drakenclimber, line 70: We probably need an `else` here to handle the case where the create_process erroneously succeeds

tests/ftests/088-sudo-cgclassify_systemd_scope.py
```diff
@@ -0,0 +1,169 @@
...
+    # Test 3: create a non-scope cgroup, try creating a task with replace_idle
+    #         set.
+    #
+    Cgroup.create_and_validate(config, CONTROLLER, OUT_OF_SCOPE_CGNAME1)
+    try:
+        config.process.create_process_in_cgroup(config, CONTROLLER, OUT_OF_SCOPE_CGNAME1,
+                                                replace_idle=True)
+    except RunError as re:
+        if "Failed to find idle_thread task" not in re.stderr:
+            raise re
```
drakenclimber, line 92: I think an `else` here would also be good

tests/ftests/088-sudo-cgclassify_systemd_scope.py
```diff
@@ -0,0 +1,169 @@
...
+    # Test 4: create a non-scope cgroup with .scope suffix, try creating a task with
+    #         replace_idle set.
+    #
+    Cgroup.create_and_validate(config, CONTROLLER, OUT_OF_SCOPE_CGNAME2)
+    try:
+        config.process.create_process_in_cgroup(config, CONTROLLER, OUT_OF_SCOPE_CGNAME2,
+                                                replace_idle=True)
+    except RunError as re:
+        if "Failed to find idle_thread task" not in re.stderr:
+            raise re
```
drakenclimber, line 114: Ditto on the `else`

src/tools/cgclassify.c
```diff
@@ -26,6 +31,18 @@
...
+
+static pid_t find_scope_pid(pid_t pid, int capture);
+static int write_systemd_unified(const char * const cgrp_name, pid_t pid);
+static int is_scope_parsed(const char * const path);
+static int rollback_pid_cgroups(pid_t pid);
+
+struct cgroup_info {
+	char ctrl_name[CONTROL_NAMELEN_MAX];
+	char cgrp_path[FILENAME_MAX];
+}info[16];
```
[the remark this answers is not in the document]
kamalesh-babulal: @drakenclimber Thank you so much for reviewing the patch set. Yeah, I miss this instance will replace it.

src/tools/cgclassify.c
kamalesh-babulal: Sure, this sounds like a good idea to add a safety check.

src/tools/cgexec.c
```diff
@@ -171,12 +190,327 @@ int main(int argc, char *argv[])
...
+	if (!replace_idle) {
+		/* Now exec the new process */
+		execvp(argv[optind], &argv[optind]);
+		err("exec failed:%s", strerror(errno));
+		return -1;
+	}
+
+	scope_pid = find_scope_pid(pid);
+	if (scope_pid == -1)
+		return -1;
```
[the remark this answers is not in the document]
kamalesh-babulal: `find_scope_pid()` prints the error message before returning `-1`, it will be redundant here.  Does that sound okay, to skip it here?

src/tools/cgexec.c
```diff
@@ -171,12 +190,327 @@ int main(int argc, char *argv[])
...
+		err("Failed to open %s\n", cgroup_procs_path);
 		return -1;
 	}
 
+	if (is_cgroup_mode_hybrid()) {
+		/*
+		 * construct the unified cgroup path, by parsing
+		 * /proc/mounts
+		 */
+		find_mnt_point("unified cgroup2", &cgroup_name);
```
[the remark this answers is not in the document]
kamalesh-babulal: ah !! good catch will fix them.

src/tools/cgexec.c
```diff
@@ -171,12 +190,327 @@ int main(int argc, char *argv[])
...
+
+static int is_scope_parsed(const char * const path)
+{
+	/*
+	 * As per, <kernel sources>/kernel/cgroup/cgroup.c::cgroup_init()
+	 * At the max there can be only 16 controllers and we are
+	 * not accounting for named hierarchies, which can be more
+	 * than 16 themselves.
+	 */
+	static char parsed_scope_path[MAX_MNT_ELEMENTS][FILENAME_MAX];
```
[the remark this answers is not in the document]
kamalesh-babulal: There are implicitly initialized to zeros, we can skip the initialization, AFAIK.

tests/ftests/088-sudo-cgclassify_systemd_scope.py
kamalesh-babulal: Good idea will add `cgclassify=True` to the argument list.

tests/ftests/088-sudo-cgclassify_systemd_scope.py
kamalesh-babulal: Make sense will add the `else` block too for erroneous success.

tests/ftests/088-sudo-cgclassify_systemd_scope.py
kamalesh-babulal: Agreed will add the `else` code.

PR #356, opened 2023-06-12
config.c: fix memory leak in convert_controller_versions()

kamalesh-babulal:
Fix memory leak warning in convert_controller_versions(), reported by Coverity tool:

CID 313911 (#1 of 1): Resource leak (RESOURCE_LEAK)9. leaked_storage:
Variable cgrp_cpy going out of scope leaks the storage it points to.

struct cgroup cgrp_cpy was not freed in the success path, fix it.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -852,7 +852,8 @@ next_controller:
 		cgroup_free_controller(cgc);
 		cgrp_cpy->index = 0;
 	}
-	return 0;
+
+	ret = 0;
 err:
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit https://github.com/libcgroup/libcgroup/commit/1b53e2af890f07f519c7074d052380f6132850f1.  Good find, @kamalesh-babulal

PR #359, opened 2023-06-16
[v2] tools/cgclassify: add -r option for systemd scope

kamalesh-babulal:
This patch set provides an option for users to replace the default idle_thread
created for systemd scope. It introduces '-r' option, when specified, the first
pid passed as the argument to cgclassify will replace the default idle_thread,
if it exists. Also, if it's the only task, killing it will remove the .scope cgroup it's
running in. It also adds support for ftests to pass the -r flag and a test case to
exercise the different likely and unlikely paths with replace flag.

### DIFF
--- a/src/tools/cgclassify.c
+++ b/src/tools/cgclassify.c
@@ -3,6 +3,10 @@
  * Copyright RedHat Inc. 2008
  *
  * Authors:	Vivek Goyal <vgoyal@redhat.com>
+ *
+ * Replace systemd idle_thread enhancements by Kamalesh Babulal
+ * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Author: Kamalesh Babulal <kamalesh.babulal@oracle.com>
  */
 
 #include "tools-common.h"
@@ -15,6 +19,7 @@
 #include <limits.h>
 #include <unistd.h>
 #include <getopt.h>
+#include <signal.h>
 #include <errno.h>
 #include <stdio.h>
 #include <pwd.h>
@@ -25,6 +30,18 @@
 
 
 #define TEMP_BUF	81
+
+#define SYSTEMD_IDLE_THREAD	"libcgroup_systemd_idle_thread"
+
+static pid_t find_scope_pid(pid_t pid, int capture);
+static int write_systemd_unified(const char * const cgrp_name, pid_t pid);
+static int is_scope_parsed(const char * const path);
+static int rollback_pid_cgroups(pid_t pid);
+
+struct cgroup_info {
+	char ctrl_name[CONTROL_NAMELEN_MAX];
+	char cgrp_path[FILENAME_MAX];
+}info[MAX_MNT_ELEMENTS];
 
 static void usage(int status, const char *program_name)
 {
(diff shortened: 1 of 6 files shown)


kamalesh-babulal:
**V2 Changes**:
- Rebased the code based on the latest main.
- Addressed the reviewed comments from @drakenclimber #355 
  - Fixed memory leaks from `find_mnt_point()` 
  - Added `else` false positive cases

src/tools/cgclassify.c
```diff
@@ -198,7 +231,459 @@ int main(int argc, char *argv[])
...
+	int err = 0, idx = 0, ret = 0;
+	char *cgrp_proc_path = NULL;
+
+	/*
+	 * unified cgroup rollback is simple, we need to write into
+	 * single cgroup hierarchy.
+	 */
+	if (is_cgroup_mode_unified()) {
+		pthread_rwlock_rdlock(&cg_mount_table_lock);
+		cg_build_path_locked(info[idx].cgrp_path, cgroup_path, NULL);
```
drakenclimber, line 638: What if a user does something like this:
```
cgcreate -c -S -g cpu:my.slice/db.scope   # set my.slice/db.scope as the default
cgclassify -r -g cpu:my.slice/db.scope 1234
```

This line would have `cg_build_path_locked()` prepend `my.slice/db.scope` in front of `info[idx].cgroup_path`.  But since `info[idx].cgroup_path` came directly from `/proc/{pid}/cgroup`, I don't see why we would ever want to prepend the slice/scope info.  What am I missing?

src/tools/cgclassify.c
```diff
@@ -198,7 +231,459 @@ int main(int argc, char *argv[])
...
+
+		snprintf(cgroup_proc_path, FILENAME_MAX + 14, "%s/cgroup.procs", cgroup_path);
+		ret = attach_task_pid(cgroup_proc_path, pid);
+		return ret;
+	}
+
+	for (idx = 0; info[idx].ctrl_name[0] != '\0'; idx++) {
+		/* find the systemd cgroup path */
+		if (!strcmp(info[idx].ctrl_name, "name=systemd")) {
+			find_mnt_point("name=systemd ", &cgrp_proc_path);
```
drakenclimber, line 649: You have a space at the end of `name=systemd `.  If a system ever places the `name=systemd` option before `noexec`, `nodev`, etc. then it won't be found.  Is this by design?

kamalesh-babulal: @drakenclimber Thank you so much for the review. Agreed, it's not robust will come up with something better.

src/tools/cgclassify.c
kamalesh-babulal: In this above example the user is expected to use one of the following commands:
`cgclassify -b -r -g cpu:my.slice/db.scope 1234` or
`cgclassify -r -g cpu:/ 1234`

either the user could use `-b` option to ignore the default delegated `slice/scope` or pass `/` referring to the root a.k.a delegated `slice and scope`

kamalesh-babulal:
**v3 Changes**
- Addressed the review comments from @drakenclimber 
- Fixed a memory leak and uninitialized var in `find_scope_pid()`
- Improved error path return.

src/tools/cgclassify.c
drakenclimber: But since the cgroup path info is coming from `/proc/{pid}/cgroup`, the slice/scope are already there right?  So, wouldn't your first example end up with a path like `/sys/fs/cgroup/my.slice/db.scope/my.slice/db.scope`?

drakenclimber: Gotcha.  I missed the `-b` in your first example. Thanks!

drakenclimber (approved):

PR #360, opened 2023-06-18
tools/cgexec: fix memory leak in find_scope_pid()

kamalesh-babulal:
Fix a memory leak, reported by the Coverity tool.

CID 321265 (#1-2 of 2): Resource leak (RESOURCE_LEAK)37.
leaked_storage: Variable pids going out of scope leaks the storage it points to.

`find_scope_pid ()`, calls `cgroup_get_procs()` to get the pids of the tasks under
cgroup name, and it's up to the caller to free the allocated pids. Fix it by calling
`free(pids)` after every call to `cgroup_get_procs()`.


drakenclimber (approved):

drakenclimber:

PR #361, opened 2023-06-20
tools/cgexec: fix uninitialized variable in find_scope_pid()

kamalesh-babulal:
Fix uninitialized variable, reported by the Coverity tool:

CID 321266: Uninitialized scalar variable (UNINIT)

`memset()`, `ctrl_name` with every iteration, to avoid using the
stale value or uninitialized values.


drakenclimber (approved):

drakenclimber:

PR #362, opened 2023-06-21
cgcreate: fix null pointer dereference in create_systemd_scope()

kamalesh-babulal:
Fix a possible NULL pointer dereference, reported by the Coverity tool:

CID 321267 (#1 of 1): Dereference null return value (NULL_RETURNS)6.
dereference: Dereferencing a pointer that might be NULL scope when
calling strlen.

check for the return value from `strstr()`, before dereferencing it.


drakenclimber (approved):
I don't think it's possible to get here (because `cgroup_create_scope2()` will fail if there's no `/`), but better safe than sorry.

drakenclimber:

PR #363, opened 2023-06-23
api: add ret value check in cgroup_get_current_controller_path()

kamalesh-babulal:
Fix unchecked return value, reported by the Coverity tool:

CID 320874 (#1 of 1): Unchecked return value from library
(CHECKED_RETURN)13. check_return: Calling fscanf(pid_cgroup_fd,
"%*[^\n]\n") without checking return value. This library function may

add the missing return value check from `fscanf()` in 
`cgroup_get_current_controller_path()`


drakenclimber (approved):

drakenclimber:

PR #366, opened 2023-06-28
fix: malloc size off-by-one in cgroup_process_v2_mnt

adriaan42:
need to allocate the length of both strings, plus the space added in the sprintf, plus the terminating null byte.


kamalesh-babulal (approved):

drakenclimber (approved):

drakenclimber:
Pushed to main.  Thanks, @adriaan42

PR #369, opened 2023-07-03
Improve test reliability

drakenclimber:
This patchset contains two small patches that should help resolve some of the recurring test failures we're seeing.
1. On underpowered systems, LXD will throw a websocket error.  In this case, wait a bit and try again
2. Test 002 sometimes fails on underpowered systems when it's run in a container.  Skip it


kamalesh-babulal (approved):

drakenclimber:

PR #370, opened 2023-07-03
Fix a few issues reported by coverity

drakenclimber:


kamalesh-babulal (approved):

drakenclimber:

PR #371, opened 2023-07-05
api: fix missing goto in cgroup_get_current_controller_path()

kamalesh-babulal:
Fix missing goto on error in `cgroup_get_current_controller_path()`,
it was reported by Coverity tool as unused value:

CID 320875 (#1 of 1): Unused value (UNUSED_VALUE)assigned_value:
Assigning value ECGOTHER to ret here, but that stored value is overwritten
before it can be used.

It turns out that, it is more than an over-written value, the `strdup()` was
missing goto error path statement on failure, add it.


drakenclimber (approved):

drakenclimber:

PR #372, opened 2023-07-06
tools/cgxget: fix unused value in main

kamalesh-babulal:
Fix unused value, reported by the Coverity tool:

CID 320879 (#1 of 1): Unused value (UNUSED_VALUE)assigned_value:
it can be used.

If the mapping from v1->v2 and vice versa fail, with `-i (ignore-unmappable)`
set, ret is set to zero and is immediately over-written, re-arrange the check,
and adjust the commentary to avoid clobbering ret, by taking the error path
when `-i` is unset or else continuing when set, skipping to set ret to zero.

### DIFF
--- a/src/tools/cgxget.c
+++ b/src/tools/cgxget.c
@@ -861,13 +861,13 @@ int main(int argc, char *argv[])
 		cgroup_set_default_systemd_cgroup();
 
 	ret = convert_cgroups(&cg_list, cg_list_len, version, CGROUP_DISK);
-	if (ret == ECGNOVERSIONCONVERT && ignore_unmappable)
+	if ((ret && ret != ECGNOVERSIONCONVERT) ||
+	    (ret == ECGNOVERSIONCONVERT && !ignore_unmappable))
 		/*
-		 * The user has specified that we should ignore any errors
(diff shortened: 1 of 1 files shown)


src/tools/cgxget.c
```diff
@@ -861,13 +861,12 @@ int main(int argc, char *argv[])
 		cgroup_set_default_systemd_cgroup();
 
 	ret = convert_cgroups(&cg_list, cg_list_len, version, CGROUP_DISK);
-	if (ret == ECGNOVERSIONCONVERT && ignore_unmappable)
+	if (ret == ECGNOVERSIONCONVERT && !ignore_unmappable)
```
drakenclimber, line 864: I don't think this is correct.  We need to fail on any value of `ret` or if `ret == ECGNOVERSIONCONVERT` and `ignore_unmappable` is not set.

What do you think of this?
```
if ((ret != ECGNOVERSIONCONVERT && ret) ||
    (ret == ECGNOVERSIONCONVERT && !ignore_unmappable))
```

drakenclimber (approved):

drakenclimber:

PR #373, opened 2023-07-07
tools/cgxset: fix unused value in main

kamalesh-babulal:
Fix unused value, reported by the Coverity tool:

CID 320876 (#1 of 1): Unused value (UNUSED_VALUE)assigned_value:
it can be used.

If the mapping from v1->v2 and vice versa fail, with `-i` (ignore-unmappable)
set, ret is set to zero and is immediately over-written, re-arrange the check,
and adjust the commentary to avoid clobbering ret, by taking the error path
when `-i` is unset or else continuing when set, skipping to set ret to zero.


drakenclimber (approved):

drakenclimber:

PR #374, opened 2023-07-08
api: cg_move_task_files() - rearrange code to avoid extra check

kamalesh-babulal:
Fix unused variables, reported by the Coverity tool:

CID 320878 (#1 of 1): Unused value (UNUSED_VALUE)assigned_value:
it can be used.

Currently, in `cg_move_task_files()`, the ret is checked for `ESRCH` and
set to zero, if true, and for other error values it executes the error path
but immediately the ret is over-written without reading the set zero,
re-arrange the code to such that only the error path is executed and
ret is not set to zero.


drakenclimber (approved):

drakenclimber:

PR #377, opened 2023-07-14
tools/cgget,cgxget: fix TOCTOU warning

kamalesh-babulal:
This  patch series fixes TOCTOU warnings in `cgget` and `cgxget`
tools, reported by the Coverity tool.  The   Coverity gets confused
when a char array is re-used for constructing different paths and
complains about TOCTOU for unrelated paths, this patch series
fixes it by using different char arrays for different paths, as a side
effect it also improves the readability of the code.


drakenclimber (approved):

drakenclimber:

PR #378, opened 2023-07-15
src/api.c: fix TOCTOU in cgroup_get_cgroup()

kamalesh-babulal:
Fix a TOCTOU issue, reported by Coverity tool:

CID 258268 (#1 of 1): Time of check time of use (TOCTOU)28.
fs_check_call: Calling function access to perform check on path.

Coverity gets confused when a char array is re-used for constructing
different paths and complains about TOCTOU for unrelated paths,
fix it by using different char arrays for different path, as a side effect
it also improves the readability of the code.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -3643,11 +3643,12 @@ fill_error:
  */
 int cgroup_get_cgroup(struct cgroup *cgroup)
 {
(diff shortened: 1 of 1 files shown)


drakenclimber:

PR #380, opened 2023-07-17
src/api.c fix TOCTOU in cg_mkdir_p()

kamalesh-babulal:
Fix a TOCTOU issue, reported by the Coverity tool:

CID 258294 (#1 of 1): Time of check time of use (TOCTOU)42.
fs_check_call: Calling function stat to perform check on real_path

Add new switch for EROFS, to explicitly check if the file exists on
`mkdir()'s` error path, using a new char pointer.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -2109,8 +2109,9 @@ int cgroup_attach_task(struct cgroup *cgroup)
 int cg_mkdir_p(const char *path)
 {
 	char *real_path = NULL;
-	int stat_ret, ret = 0;
+	char *tmp_path = NULL;
 	struct stat st;
+	int ret = 0;
 	int i = 0;
 	char pos;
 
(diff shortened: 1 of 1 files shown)


src/api.c
```diff
@@ -2144,19 +2144,21 @@ int cg_mkdir_p(const char *path)
...
-			default:
-				/* Check if path exists */
-				real_path[i] = '\0';
-				stat_ret = stat(real_path, &st);
-				real_path[i] = pos;
-				if (stat_ret == 0) {
-					ret = 0;	/* Path exists */
-					break;
-				}
+			case EROFS:
```
kamalesh-babulal, line 2147: The  reason for explicitly checking for `EROFS` errno is  captured in the commit 74b3eb1db015

drakenclimber:

PR #381, opened 2023-07-18
python: Rename __set_default_systemd_cgroup()

drakenclimber:
Rename __set_default_systemd_cgroup() to _set_default_systemd_cgroup() to avoid Python name mangling


kamalesh-babulal (approved):

drakenclimber:
Pushed to main.  Thanks for the review, @kamalesh-babulal

PR #383, opened 2023-07-19
tools/cgsnapshot: fix string null terminate in parse_controllers()

kamalesh-babulal:
Fix string not null terminated warning, reported by the Coverity tool:

CID 258299 (#2 of 2): String not null terminated (STRING_NULL)18.
string_null: Passing unterminated string *controllers to display_controller_data,
which expects a null-terminated string.

use `snprintf()` instead of `strncpy()` and manually terminate the string
to keep Coverity happy.


drakenclimber (approved):

drakenclimber:

PR #385, opened 2023-07-19
cgsnapshot: Fix compiler truncation warning

drakenclimber:
Fix the following warning when running `make distcheck`

```
	../../../../src/tools/cgsnapshot.c: In function ‘is_ctlr_on_list’:
	../../../../src/tools/cgsnapshot.c:532:57: warning: ‘%s’ directive output may be truncated writing up to 409599 bytes into a region of size 4096 [-Wformat-truncation=]
	  532 |                 snprintf(controllers[i], FILENAME_MAX, "%s", tmp_controllers[i]);
	      |                                                         ^~
```

### DIFF
--- a/src/tools/cgsnapshot.c
+++ b/src/tools/cgsnapshot.c
@@ -529,7 +529,12 @@ static int is_ctlr_on_list(char controllers[CG_CONTROLLER_MAX][FILENAME_MAX],
 
 	/* Lets reset the controllers to intersection of controller ∩ wanted_conts */
 	for (i = 0; tmp_controllers[i][0] != '\0'; i++) {
-		snprintf(controllers[i], FILENAME_MAX, "%s", tmp_controllers[i]);
+		/*
+		 * gcc complains about truncation when using snprintf() and
+		 * and coverity complains about truncation when using strncpy().
+		 * Avoid both these warnings by directly invoking memcpy()
+		 */
+		memcpy(controllers[i], tmp_controllers[i], sizeof(controllers[i]));
 		ret = 1;
 	}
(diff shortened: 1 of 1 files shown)


src/tools/cgsnapshot.c
```diff
@@ -529,7 +529,12 @@ static int is_ctlr_on_list(char controllers[CG_CONTROLLER_MAX][FILENAME_MAX],
 
 	/* Lets reset the controllers to intersection of controller ∩ wanted_conts */
 	for (i = 0; tmp_controllers[i][0] != '\0'; i++) {
-		snprintf(controllers[i], FILENAME_MAX, "%s", tmp_controllers[i]);
+		int pret;
+
+		pret = snprintf(controllers[i], FILENAME_MAX, "%s", tmp_controllers[i]);
+		if (pret >= FILENAME_MAX)
+			cgroup_err("Warning truncation may have occurred: %s\n",
+				   controllers[i]);
```
kamalesh-babulal, line 537: @drakenclimber Sorry, I had run build against both `gcc-11` and `gcc-10`  but it did not show up because `-Wformat-truncation=` is not enabled by default.  Do you think we could use `memcpy()` here, instead of `snprintf()`  both `Coverity` and `gcc` seems to be happy with it.

drakenclimber: No worries.

I don't have a strong preference of `memcpy()` vs `snprintf()`.  It would probably be good to add a comment about the truncation warning if we go with something non-standard like `memcpy()`

drakenclimber:

Issue #387, opened 2023-07-29
3.1.0: build fails becasuse missing checking `libsystemd` pkgconfig dependency checking


kloczek:
Looks like build fails because there is no checking over pkgconfig `libsystemd` checking.
autoconf output:
```console
e --infodir=/usr/share/info --libdir=/usr/lib64 --libexecdir=/usr/libexec --localstatedir=/var --mandir=/usr/share/man --prefix=/usr --program-prefix= --runstatedir=/run --sbindir=/usr/sbin --sharedstatedir=/var/lib --sysconfdir=/etc --disable-daemon --disable-static --enable-tests --enable-opaque-hierarchy=name=systemd --enable-pam-module-dir=/usr/lib64/security
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a race-free mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether make supports nested variables... (cached) yes
checking whether make supports the include directive... yes (GNU style)
checking for x86_64-redhat-linux-gnu-gcc... /usr/bin/gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C... yes
checking whether /usr/bin/gcc accepts -g... yes
checking for /usr/bin/gcc option to enable C11 features... none needed
checking whether /usr/bin/gcc understands -c and -o together... yes
checking dependency style of /usr/bin/gcc... none
checking for x86_64-redhat-linux-gnu-ar... /usr/bin/gcc-ar
checking the archiver (/usr/bin/gcc-ar) interface... ar
checking build system type... x86_64-redhat-linux-gnu
checking host system type... x86_64-redhat-linux-gnu
checking how to print strings... printf
checking for a sed that does not truncate output... /usr/bin/sed
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for fgrep... /usr/bin/grep -F
checking for ld used by /usr/bin/gcc... /usr/bin/ld
checking if the linker (/usr/bin/ld) is GNU ld... yes
checking for BSD- or MS-compatible name lister (nm)... /usr/bin/gcc-nm
checking the name lister (/usr/bin/gcc-nm) interface... BSD nm
[...]

PR #388, opened 2023-07-31
configure.ac: add check for systemd library and header files

kamalesh-babulal:
systemd support is now enabled by default, unless explicitly disabled
during the `./configure` by passing `--enable-systemd=no`. The make
may fail with:
```
systemd.c:9:10: fatal error: systemd/sd-bus.h: No such file or directory
    9 | #include <systemd/sd-bus.h>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```
if the host system is missing the systemd library and development files,
due to missing checks for their existence in `./configure.ac`. This patch add them.

Fixes: https://github.com/libcgroup/libcgroup/issues/387
Reported-by: Tomasz Kłoczko <kloczek@github.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -236,6 +236,20 @@ if test x$with_pam = xtrue; then
 		header files!])])
 fi
 
+if test x$with_systemd = xtrue; then
+       AC_CHECK_LIB([systemd],
+                    [sd_bus_message_new_method_call],
+                    [],
+                    [AC_MSG_ERROR([Cannot compile systemd support without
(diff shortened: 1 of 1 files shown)


drakenclimber:
@kamalesh-babulal, I've created pull request #389.  It includes this fix, and I've added a few other patches.  Ultimately that patchset will use Github Actions to build/test a libcgroup build with systemd fully disabled.  Let me know what you think :+1:

PR #389, opened 2023-08-03
Add a Github Actions Job to build and test w/ systemd fully disabled

drakenclimber:


kamalesh-babulal:
@drakenclimber Interesting idea with the `m4`, thanks for working on it.

drakenclimber:
> @drakenclimber Interesting idea with the `m4`, thanks for working on it.

@kamalesh-babulal - what do you think about the weird escaping that I had to do in the *.pyx files?  `format` and `len` are keywords in m4 as well and needed to be escaped.  Also, any comma in a Python comment needed to escaped.

These aren't showstoppers for me, but we'll definitely need to be cognizant of them as we develop Python bindings.  We should make sure to review the Python output by m4 to ensure it doesn't do anything strange.

kamalesh-babulal:
> what do you think about the weird escaping that I had to do in the *.pyx files? format and len are keywords in m4 as well and needed to be escaped. Also, any comma in a Python comment needed to escaped - but commas in regular code didn't need escaping. Strange.

I did notice it but thought like every language has its keywords, `m4` complaint about the ones you had escaped.

>These aren't showstoppers for me, but we'll definitely need to be cognizant of them as we develop Python bindings. We should make sure to review the Python output by m4 to ensure it doesn't do anything strange.

Agreed, we will learn more about them, as we progress in this direction. I am sure most of the cases with will caught by the test cases.

drakenclimber:
Pushed to the main and release-3.1 branches.  Thanks for the help, @kamalesh-babulal

PR #391, opened 2023-08-07
checkpatch improvements

kamalesh-babulal:
This patch series adds two patches, that improve the `checkpatch.pl`
checks by customizing the warnings we care about.  The first patch
add more `--ignore` types for string functions, which `checkpatch.pl`
complains about, assuming it is checking against the kernel source
and the second patch introduces `ARRAY_SIZE()` macro to calculate
the array size, as suggested by the `checkpatch.pl`, which will be
used by future patches.


drakenclimber:
Pushed to main only.  Thanks, @kamalesh-babulal

PR #392, opened 2023-08-08
Use ARRAY_SIZE macro helper across source

kamalesh-babulal:
This patch series uses `ARRAY_SIZE()` macro helper to calculate array
size across the source tree, including the test cases.


drakenclimber (approved):

drakenclimber:
Pushed to the main branch.  I did _not_ push it to the release-3.1 branch.  Thanks, @kamalesh-babulal

PR #395, opened 2023-08-18
systemd check for controllers name in scope name

kamalesh-babulal:
systemd will silently prefix a '_' to the scope name and create, and
delegate it under the slice. If it matches with any of the original
cgroup and pseudo BPF-base systemd controllers. i.e.,
    
```
# cgcreate -c -g cpuset,cpu:oracle.slice/cpuset.scope
# tree /sys/fs/cgroup/oracle.slice/ -d
/sys/fs/cgroup/oracle.slice/
├── _cpuset.scope
└── cpuset.scope
    
2 directories
    
# systemd-cgls /oracle.slice
Control group /oracle.slice:
└─cpuset.scope …
  └─969 libcgroup_systemd_idle_thread
    
# cat /proc/969/cgroup
0::/oracle.slice/_cpuset.scope
```    
this implicit rename will cause confusion to the users, who would not
see any errors during creation but operate on non-delegated cgroup scope
created by libcgroup internally. 

The first patch in the patchset adds checks to disallow such systemd
scope names and the second patch updates the simple scope creation
test case to validate scope name with controller name, i.e., `cpu.scope`

### DIFF
--- a/src/systemd.c
+++ b/src/systemd.c
@@ -26,6 +26,29 @@ static const char * const modes[] = {
 static_assert(ARRAY_SIZE(modes) == CGROUP_SYSTEMD_MODE_CNT,
 	      "modes[] array must be same length as CGROUP_SYSTEMD_MODE_CNT");
 
+/*
+ * controller_name table borrowed from:
+ * https://github.com/systemd/systemd/blob/main/src/basic/cgroup-util.c#L2260
+ */
+static const char * const controller_name[] = {
+	/* subset of controllers (cgroup v1/v2), systemd supports */
+	"cpu",
+	"cpuacct",
+	"cpuset",
+	"io",
+	"blkio",
+	"memory",
+	"devices",
+	"pids",
+
+	/* systemd pseudo BPF based controllers (cgroup v2 ) */
+	"bpf-firewall",
+	"bpf-devices",
+	"bpf-foreign",
+	"bpf-socket-bind",
+	"bpf-restrict-network-interfaces",
+};
+
 static const char * const sender = "org.freedesktop.systemd1";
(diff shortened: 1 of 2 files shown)


drakenclimber:
Here's the relevant systemd code that adds the underscore prefix:
https://github.com/systemd/systemd/blob/main/src/basic/cgroup-util.c#L1557

src/systemd.c
```diff
@@ -84,6 +107,55 @@ static int job_removed_callback(sd_bus_message *message, void *user_data, sd_bus
 	return 0;
 }
 
+static int validate_scope_slice_name(const char * const scope_name, const char * const slice_name)
+{
+	char *_scope_name = NULL;
+	int i, len, ret = 0;
+
+	len = strlen(scope_name) - 5;	/* strlen(".scope") */
```
drakenclimber, line 115: I think I'm guilty of this bug.  Do we need to check for underflow here?  What if the scope is "foo" or the slice is "bar"?

kamalesh-babulal: @drakenclimber Thanks for the review. Sorry, I should have looked, I will spin v2 with the check for length.

drakenclimber: No need to apologize since I'm the one who originally wrote it :/.  Thanks for the help!

kamalesh-babulal:
**V2 Changes**
- Add underflow checks for the `scope` and `slice` name (recommend by @drakenclimber)

drakenclimber (approved):

drakenclimber:
Pushed to the main branch.  I did _not_ push it to the release-3.1 branch.

Thanks, @kamalesh-babulal

PR #397, opened 2023-09-01
.checkpatch.conf: add FILE_PATH_CHANGES to ignore list

kamalesh-babulal:
Add `FILE_PATH_CHANGES` to the list of warnings, that is very specific
to Linux Kernel, which is triggered when a file is added, moved, or 
deleted to the source tree. Let's ignore it and this patch also removes
an extra new line in the `.checkpatch.conf`.

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -18,7 +18,10 @@
 # introducing it and remove it.
 --ignore NEW_TYPEDEFS
 
-
 # Ignore Kernel str* function recommendations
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main.  I did not push it to the release-3.1 branch since the checkpatch Github Actions action isn't there.  Thanks, @kamalesh-babulal

PR #398, opened 2023-09-04
[RFC] match naming cgrp with upstream Linux

kamalesh-babulal:
This patchset renames the local variable `cgroup` -> `cgrp` to match
upstream Linux Kernel, across the files (most of the occurrence)
bringing it closer to the Linux kernel cgroup code.

### DIFF
--- a/include/libcgroup/config.h
+++ b/include/libcgroup/config.h
@@ -100,8 +100,7 @@ int cgroup_load_templates_cache_from_files(int *file_index);
  * @param tmpl_files
  */
 struct cgroup_string_list;
-void cgroup_templates_cache_set_source_files(
-	struct cgroup_string_list *tmpl_files);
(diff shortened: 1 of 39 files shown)


fmoessbauer:
This is only an internal change, right? Then I like it as it resolves the ambiguity between the parameter types and values.

kamalesh-babulal:
@fmoessbauer Thanks for looking through it. Yes, the renaming is limited to internal code.

kamalesh-babulal:
**V2 Changes**:
- Rebased to the current `main` branch

drakenclimber:
Pushed to main.  Thanks for your patience, @kamalesh-babulal

kamalesh-babulal:
@drakenclimber Thank you so much for reviewing and merging it.

PR #402, opened 2023-09-08
tools/cgset: add -R option to recursively set variables

kamalesh-babulal:
Add `-R` option to recursively set variable(s) passed to cgroups under `<cgroup_path>`. 
This will help users to set a controller setting for all the cgroups under a cgroup hierarchy,
instead of passing the cgroups multiple times on the command line.
```
example:
--------
./cgcreate -gcpu,memory:foo -gcpu:memory:foo/ch1 -gcpu,memory:foo/ch2 ./cgget -r cpu.shares foo foo/ch1 foo/ch2
foo:
cpu.shares: 1024

foo/ch1:
cpu.shares: 1024

foo/ch2:
cpu.shares: 1024

Without the patch
------------------
./cgset -r cpu.shares=256 foo
./cgget -r cpu.shares foo
foo/ch1 foo/ch2
foo:
cpu.shares: 256

foo/ch1:
cpu.shares: 1024

foo/ch2:
cpu.shares: 1024

With the patch
--------------
./cgset -R -r cpu.shares=512 foo
./cgget  -r cpu.shares foo foo/ch1 foo/ch2
foo:
cpu.shares: 512

foo/ch1:
cpu.shares: 512

foo/ch2:
cpu.shares: 512
```

### DIFF
--- a/src/tools/cgset.c
+++ b/src/tools/cgset.c
@@ -27,6 +27,8 @@ static const struct option long_options[] = {
 
 int flags; /* used input method */
 
+static char *program_name;
+
 static struct cgroup *copy_name_value_from_cgroup(char src_cg_path[FILENAME_MAX])
 {
 	struct cgroup *src_cgroup;
@@ -54,7 +56,38 @@ scgroup_err:
 	return NULL;
 }
 
-static void usage(int status, const char *program_name)
+static int cgroup_set_cgroup_values(struct cgroup *src_cgrp, const char * const new_cgrp)
+{
+	struct cgroup *cgrp;
+	int ret = 0;
+
+	/* create new cgroup */
+	cgrp = cgroup_new_cgroup(new_cgrp);
+	if (!cgrp) {
+		ret = ECGFAIL;
+		err("%s: can't add new cgroup: %s\n", program_name, cgroup_strerror(ret));
+		return ret;
+	}
+
+	/* copy the values from the source cgroup to new one */
+	ret = cgroup_copy_cgroup(cgrp, src_cgrp);
+	if (ret != 0) {
+		err("%s: cgroup %s error: %s\n", program_name, src_cgrp->name,
+		    cgroup_strerror(ret));
+		goto err;
+	}
+
+	/* modify cgroup based on values of the new one */
+	ret = cgroup_modify_cgroup(cgrp);
+	if (ret)
+		err("%s: cgroup modify error: %s\n", program_name, cgroup_strerror(ret));
+
+err:
+	cgroup_free(&cgrp);
+	return ret;
+}
+
+static void usage(int status)
 {
 	if (status != 0) {
 		err("Wrong input parameters, ");
@@ -139,14 +172,15 @@ int main(int argc, char *argv[])
 
 	char src_cg_path[FILENAME_MAX] = "\0";
 	struct cgroup *src_cgroup = NULL;
-	struct cgroup *cgroup = NULL;
 
 	int ret = 0;
 	int c;
 
+	program_name = argv[0];
+
 	/* no parameter on input */
 	if (argc < 2) {
-		err("Usage is %s -r <name=value> relative path to cgroup>\n", argv[0]);
+		err("Usage is %s -r <name=value> relative path to cgroup>\n", program_name);
 		exit(EXIT_BADARGS);
 	}
 
@@ -162,13 +196,13 @@ int main(int argc, char *argv[])
 		switch (c) {
 #endif
 		case 'h':
-			usage(0, argv[0]);
+			usage(0);
 			ret = 0;
 			goto err;
 
 		case 'r':
 			if ((flags &  FL_COPY) != 0) {
-				usage(1, argv[0]);
+				usage(1);
 				ret = EXIT_BADARGS;
 				goto err;
 			}
@@ -180,13 +214,13 @@ int main(int argc, char *argv[])
 				name_value = (struct control_value *)
 					realloc(name_value, nv_max * sizeof(struct control_value));
 				if (!name_value) {
-					err("%s: not enough memory\n", argv[0]);
+					err("%s: not enough memory\n", program_name);
 					ret = -1;
 					goto err;
 				}
 			}
 
-			ret = parse_r_flag(argv[0], optarg, &name_value[nv_number]);
+			ret = parse_r_flag(program_name, optarg, &name_value[nv_number]);
 			if (ret)
 				goto err;
 
@@ -194,7 +228,7 @@ int main(int argc, char *argv[])
 			break;
 		case COPY_FROM_OPTION:
 			if (flags != 0) {
-				usage(1, argv[0]);
+				usage(1);
 				ret = EXIT_BADARGS;
 				goto err;
 			}
@@ -203,7 +237,7 @@ int main(int argc, char *argv[])
 			src_cg_path[FILENAME_MAX-1] = '\0';
 			break;
 		default:
-			usage(1, argv[0]);
+			usage(1);
 			ret = EXIT_BADARGS;
 			goto err;
 		}
@@ -211,13 +245,13 @@ int main(int argc, char *argv[])
 
 	/* no cgroup name */
 	if (!argv[optind]) {
-		err("%s: no cgroup specified\n", argv[0]);
+		err("%s: no cgroup specified\n", program_name);
 		ret = EXIT_BADARGS;
 		goto err;
 	}
 
 	if (flags == 0) {
-		err("%s: no name-value pair was set\n", argv[0]);
+		err("%s: no name-value pair was set\n", program_name);
 		ret = EXIT_BADARGS;
 		goto err;
 	}
@@ -225,7 +259,8 @@ int main(int argc, char *argv[])
 	/* initialize libcgroup */
 	ret = cgroup_init();
 	if (ret) {
-		err("%s: libcgroup initialization failed: %s\n", argv[0], cgroup_strerror(ret));
+		err("%s: libcgroup initialization failed: %s\n", program_name,
+		    cgroup_strerror(ret));
 		goto err;
 	}
 
@@ -248,39 +283,15 @@ int main(int argc, char *argv[])
 	}
 
 	while (optind < argc) {
-		/* create new cgroup */
-		cgroup = cgroup_new_cgroup(argv[optind]);
-		if (!cgroup) {
-			ret = ECGFAIL;
-			err("%s: can't add new cgroup: %s\n", argv[0], cgroup_strerror(ret));
-			goto cgroup_free_err;
-		}
-
-		/* copy the values from the source cgroup to new one */
-		ret = cgroup_copy_cgroup(cgroup, src_cgroup);
-		if (ret != 0) {
-			err("%s: cgroup %s error: %s\n", argv[0], src_cg_path,
-			    cgroup_strerror(ret));
-			goto cgroup_free_err;
-		}
-
-		/* modify cgroup based on values of the new one */
-		ret = cgroup_modify_cgroup(cgroup);
-		if (ret) {
-			err("%s: cgroup modify error: %s\n", argv[0], cgroup_strerror(ret));
-			goto cgroup_free_err;
-		}
+		ret = cgroup_set_cgroup_values(src_cgroup, argv[optind]);
+		if (ret)
+			goto err;
 
 		optind++;
-		cgroup_free(&cgroup);
 	}
 
-cgroup_free_err:
-	if (cgroup)
-		cgroup_free(&cgroup);
-	if (src_cgroup)
-		cgroup_free(&src_cgroup);
 err:
+	cgroup_free(&src_cgroup);
 	free(name_value);
 
 	return ret;
--- a/src/tools/cgset.c
+++ b/src/tools/cgset.c
@@ -9,6 +9,7 @@
 #include <unistd.h>
 #include <getopt.h>
 #include <stdio.h>
+#include <errno.h>
 
 #define FL_RULES	1
 #define FL_COPY		2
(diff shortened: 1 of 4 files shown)


kamalesh-babulal:
**V2 changes**:
- Added man page to document `-R` option.

kamalesh-babulal:
After the review,  I will work on the test cases and python API.

kamalesh-babulal:
**v3**:
- Fix the issue with the leaf node for cgroup v2
- Added support to set controller setting recursively for `set()` in ftests/cgroup
- Added a test case to test cgset `-R` option.

drakenclimber:
Out of curiosity, I tested how it behaved with `cgroup.subtree_control` only enabled for the `cpu` controller on some of the children and grandchildren.

```
cd /sys/fs/cgroup
mkdir parent
mkdir parent/child1
mkdir parent/child1/grandchild1
mkdir parent/child1/grandchild2
mkdir parent/child2

echo +cpu > parent.cgroup.subtree_control
```

```
sudo ./src/tools/cgset -R -r cpu.weight=101 parent
...
parent/child1/grandchild1: cgroup modify error: Cgroup, requested group parameter does not exist

find parent/ -name cpu.weight | xargs cat
101
101
100
```

It behaved as I hoped.  Some of the `cpu.weight` values were changed, but it puked when I expected.  Any `cpu.weight` values after that didn't get changed.

src/tools/cgset.c
```diff
@@ -54,6 +56,97 @@ static struct cgroup *copy_name_value_from_cgroup(char src_cg_path[FILENAME_MAX]
...
+
+err:
+
+	if (ret == ECGEOF) {
+	/* we successfully walked the tree */
+		ret = 0;
+
+		if (!is_cgroup_mode_unified()) {
+			ret = modify_cgroup(src_cgrp, &parent_info_full_path[prefix_len]);
+			free(parent_info_full_path);
```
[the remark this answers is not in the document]
kamalesh-babulal: @drakenclimber Thank you so much for the review. Yes, this is  memory  leak, will fix it in the  next iteration.

src/tools/cgset.c
```diff
@@ -248,39 +346,23 @@ int main(int argc, char *argv[])
...
+				goto err;
 		}
 
-		/* modify cgroup based on values of the new one */
-		ret = cgroup_modify_cgroup(cgroup);
-		if (ret) {
-			err("%s: cgroup modify error: %s\n", argv[0], cgroup_strerror(ret));
-			goto cgroup_free_err;
-		}
+		ret = modify_cgroup(src_cgroup,  argv[optind]);
```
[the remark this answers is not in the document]
kamalesh-babulal: Good  catch, yes it  missing the `else` block. Will fix this too in the next  iteration.

kamalesh-babulal:
>  It behaved as I hoped. Some of the cpu.weight values were changed, but it puked when I expected. Any cpu.weight values after that didn't get changed.

Thank you for testing,  it exposes an issue with the logic written for `unified` mode, in case of N leaf nodes. I will work on the v3 and post  it soon.

kamalesh-babulal:
**v4 changes**:
- Improved the logic to handle N  leaf nodes (pointed  from @drakenclimber)
- Fixed  review comments  from @drakenclimber

kamalesh-babulal:
**V5 changes**:
- Fixed the `cgset` man page.

drakenclimber:
I'm not sure the leaf node logic is behaving as I expected.  @kamalesh-babulal could you help me out here?

I created an arbitrary hierarchy:
```
# find /sys/fs/cgroup/grandparent/ -type d | sort
/sys/fs/cgroup/grandparent/
/sys/fs/cgroup/grandparent/parent1
/sys/fs/cgroup/grandparent/parent1/child1
/sys/fs/cgroup/grandparent/parent1/child2
/sys/fs/cgroup/grandparent/parent2
/sys/fs/cgroup/grandparent/parent2/child3
/sys/fs/cgroup/grandparent/parent2/child4
/sys/fs/cgroup/grandparent/parent2/child5
/sys/fs/cgroup/grandparent/parent3
/sys/fs/cgroup/grandparent/parent3/child6
/sys/fs/cgroup/grandparent/parent3/child7
```

I then enabled the cpu controller in grandparent, parent1, and parent3.  Notice that parent2's children don't have cpu.weight settings.  Here's the cpu.weight settings I have available in my hierarchy:

```
# find /sys/fs/cgroup/grandparent/ -name cpu.weight | sort
/sys/fs/cgroup/grandparent/cpu.weight
/sys/fs/cgroup/grandparent/parent1/child1/cpu.weight
/sys/fs/cgroup/grandparent/parent1/child2/cpu.weight
/sys/fs/cgroup/grandparent/parent1/cpu.weight
/sys/fs/cgroup/grandparent/parent2/cpu.weight
/sys/fs/cgroup/grandparent/parent3/child6/cpu.weight
/sys/fs/cgroup/grandparent/parent3/child7/cpu.weight
/sys/fs/cgroup/grandparent/parent3/cpu.weight
```

I then ran the following command
```
# sudo ./src/tools/cgset -R -r cpu.weight=101 grandparent;echo $?
0
```

Here's the results of the operation:
```
/sys/fs/cgroup/grandparent/cpu.weight = 101
/sys/fs/cgroup/grandparent/parent1/cpu.weight = 101
/sys/fs/cgroup/grandparent/parent1/child1/cpu.weight = 100
/sys/fs/cgroup/grandparent/parent1/child2/cpu.weight = 100
/sys/fs/cgroup/grandparent/parent2/cpu.weight = 101
cat: /sys/fs/cgroup/grandparent/parent2/child3/cpu.weight: No such file or directory
/sys/fs/cgroup/grandparent/parent2/child3/cpu.weight =
cat: /sys/fs/cgroup/grandparent/parent2/child4/cpu.weight: No such file or directory
/sys/fs/cgroup/grandparent/parent2/child4/cpu.weight =
cat: /sys/fs/cgroup/grandparent/parent2/child5/cpu.weight: No such file or directory
/sys/fs/cgroup/grandparent/parent2/child5/cpu.weight =
/sys/fs/cgroup/grandparent/parent3/cpu.weight = 101
/sys/fs/cgroup/grandparent/parent3/child6/cpu.weight = 100
/sys/fs/cgroup/grandparent/parent3/child7/cpu.weight = 100
```

It successfully ignored parent2's children and their lack of cpu.weight.  (I'm not certain if this is a good thing, honestly, but that's a separate discussion.)

But why didn't it set cpu.weight in the leaf children?  I would expect a recursive command to set values in all of its children regardless of them being leaf nodes or not.  What am I missing?

drakenclimber:
I think another interesting use case could be:

```
# sudo ./src/tools/cgset -R -r cgroup.subtree_control=+cpu grandparent;echo $?
```

We would have to ensure that we walk the tree at the grandparent level first, then parent, then children, ....

kamalesh-babulal:
> But why didn't it set cpu.weight in the leaf children? I would expect a recursive command to set values in all of its children regardless of them being leaf nodes or not. What am I missing?

@drakenclimber Thank you so much for trying out the patch.  By design, the cgroup v2 leaf nodes are not supposed to have any controller enabled, it the similar to what we do with `cgcreate`, for example:
```
# ./cgcreate -gcpu:grandparent/parent1/child1
# ./cgcreate -gcpu:grandparent/parent2
# ./cgget -r cgroup.subtree_control grandparent
grandparent:
cgroup.subtree_control: cpu

# ./cgget -r cgroup.subtree_control grandparent/parent1
grandparent/parent1:
cgroup.subtree_control: cpu

# ./cgget -r cgroup.subtree_control grandparent/parent1/child1
grandparent/parent1/child1:
cgroup.subtree_control: 

# ./cgget -r cgroup.subtree_control grandparent/parent2       
grandparent/parent2:
cgroup.subtree_control: 

```
> `# sudo ./src/tools/cgset -R -r cgroup.subtree_control=+cpu grandparent;echo $?`

This should be work with  the principal of no controllers  in the leaf node.
```
# ./cgcreate -g:grandparent/parent1/child1
# ./cgcreate -g:grandparent/parent2
# ./cgget -r cgroup.subtree_control grandparent/parent2
grandparent/parent2:
cgroup.subtree_control: 

# ./cgget -r cgroup.subtree_control grandparent/parent1
grandparent/parent1:
cgroup.subtree_control: 

# ./cgset -R -r cgroup.subtree_control=+cpu grandparent;echo $?
0
# ./cgget -r cgroup.subtree_control grandparent/
grandparent/:
cgroup.subtree_control: cpu

# ./cgget -r cgroup.subtree_control grandparent/parent1
grandparent/parent1:
cgroup.subtree_control: cpu

# ./cgget -r cgroup.subtree_control grandparent/parent2
grandparent/parent2:
cgroup.subtree_control: 

# ./cgget -r cgroup.subtree_control grandparent/parent1/child1
grandparent/parent1/child1:
cgroup.subtree_control: 

```
> We would have to ensure that we walk the tree at the grandparent level first, then parent, then children, ....

The cgroup walk order is pre-order, the walk should be something like:
```
grandparent/
grandparent/parent1/
grandparent/parent1/child1
grandparent/parent2/
```

Do  you think, we should  be enabling the setting  passed with  `-R -r <setting>=<value>` until the leaf  node for the `unified`  cgroup mode?

drakenclimber:
> > But why didn't it set cpu.weight in the leaf children? I would expect a recursive command to set values in all of its children regardless of them being leaf nodes or not. What am I missing?
> 
> @drakenclimber Thank you so much for trying out the patch. By design, the cgroup v2 leaf nodes are not supposed to have any controller enabled, it the similar to what we do with `cgcreate`, for example:

Agreed, but I wasn't enabling/disabling leaf nodes (via cgroup.subtree_control).  I was writing `cpu.weight`.  As a user, I would expect that to propagate through the entire tree, as it has no relation to leaf nodes.

kamalesh-babulal:
> Agreed, but I wasn't enabling/disabling leaf nodes (via cgroup.subtree_control). I was writing cpu.weight. As a user, I would expect that to propagate through the entire tree, as it has no relation to leaf nodes.

Sorry, re-reading your previous comment.  Agreed, I was thinking from the writing controllers into the `cgroup.sbtree_control`.  Yes, `cpu.weight` setting should be propagated until the leaf node.  Consider a case where the user wants disable a controller in the hierarchy, this becomes tricky, so we have two options:
- We disallow any `cgroup.*` (cgroup controller) modification under the `-R` flag or
- We will try writing the value in-order and the user should be mindful of what they are trying to modify?

kamalesh-babulal:
**v6 changes**:
- Rewrote the major bits of the recursive function to support v1/v2
- Improved the ftest test case.

src/tools/cgset.c
```diff
@@ -163,17 +219,85 @@ static int cgroup_copy_controller_idx(struct cgroup *dst_cgrp, struct cgroup *sr
 	return 0;
 }
 
+static int cgroup_populate_cgroup_ctrl(struct cgroup *dst, const char * const new_cgrp)
```
drakenclimber, line 222: The caller passes a NULL pointer into `dst`, and `dst` is not returned to the caller.  (I fell into this rabbit hole because I was afraid of a memory leak from the caller of this function.)

Wouldn't it be simpler/better if `dst` were a local variable:
```
static int cgroup_populate_cgroup_ctrl(const char * const new_cgrp)
{
...
        struct cgroup *dst;
....
```

src/tools/cgset.c
```diff
@@ -293,6 +417,25 @@ STATIC int parse_r_flag(const char * const program_name, const char * const name
 }
 
 #ifndef UNIT_TEST
+static int add_subtree_control_name_value(struct control_value *name_value)
+{
+	if (cgrp_subtree_ctrl_val) {
+		err("%s: duplicate -r %s option found\n", program_name, name_value->name);
+		return -1;
```
drakenclimber, line 424: It feels weird to return -1 on error here.  Would it be better to use `ECG<WHATEVER>` instead?

src/tools/cgset.c
```diff
@@ -77,11 +77,139 @@ static int cgroup_set_cgroup_values(struct cgroup *src_cgrp, const char * const
...
+
+	return 0;
+}
+
+static int cgroup_set_cgroup_values_r(struct cgroup *src_cgrp, const char * const new_cgrp)
+{
+	struct cgroup *cgrp;
+	int i, ret = ECGFAIL;
+
+	if (is_cgroup_mode_unified()) {
```
drakenclimber, line 286: What about hybrid mode?

tests/ftests/089-cgset-recursive_flag.py
```diff
@@ -0,0 +1,143 @@
...
+    Cgroup.get_and_validate(config, PARENT, SETTING_V2, DEFAULT_VALUE_V2)
+    Cgroup.get_and_validate(config, CHILD, SETTING_V2, VALUE_V2)
+    Cgroup.get_and_validate(config, GRANDCHILD, SETTING_V2, VALUE_V2)
+
+    return result, cause
+
+
+def test(config):
+
+    if Cgroup.get_cgroup_mode(config) == Mode.CGROUP_MODE_UNIFIED:
```
drakenclimber, line 115: I hate to ask this, but again what about hybrid mode?  Especially a configuration where some controllers are v1 and some are v2.

drakenclimber:
This version worked well for me on v2.  I have some concerns in the back of my head about this on a hybrid cgroup system, but I haven't tested it there.

src/tools/cgset.c
kamalesh-babulal: @drakenclimber Thank you so much for the review.  Agree, will remove it from the argument list.

src/tools/cgset.c
kamalesh-babulal: Yeah, we need to take in account the `hybrid()` mode too, will fix it in the next version.

src/tools/cgset.c
kamalesh-babulal: ah ! will change to `ECGFAIL`

tests/ftests/089-cgset-recursive_flag.py
kamalesh-babulal: @drakenclimber Sure will add the support for the `hybrid` mode too.

kamalesh-babulal:
**v7 Changes**:
- Added support for cgroup hybrid mode, recommend by @drakenclimber 
- Fixed review comments from @drakenclimber

drakenclimber (approved):

Issue #405, opened 2023-10-18
cgcreate fails to create nested cgroups for hugetlb and misc controllers (v2)


kamalesh-babulal:
`cgcreate` fails to nested cgroups (v2) when passing any one or both `hugetlb` or `misc` controllers.
```
# cgcreate -ghugetlb,misc:foo/bar
Error: Failed to delete foo/bar: No such file or directory
cgcreate: can't create cgroup foo/bar: No such file or directory
```

The reason is that `systemd` only enables  `CGROUP_MASK_V2` controller in `root_cgroup.subtree_control` by default.  The `CGROUP_MASK_V2`  has `cpu, cpuset, io, memory, pids` controllers set.  The workaround for the issue is to split the above operation into two steps:
```
# cgcreate -ghugetlb,misc:foo
# cgcreate -ghugetlb,misc:foo/bar
```
Now the controller is enabled in the parent of the cgroup `foo`, that is `/sys/fs/cgroup/cgroup.subtree_control`, and is available to all the descendant cgroups from here on.

kamalesh-babulal:
I am working on the idea of enabling it in the `root_cgroup.subtree_control` every time a controller is enabled in the cgroup. Will post the patches soon.

PR #406, opened 2023-10-19
api.c: enable controller in the root_cgroup.subtree_file

kamalesh-babulal:
`systemd` by default only enables `cpu, cpuset, io, memory`, and `pids`
controller the `root_cgroup.subtree_control`. Trying to enable other
controllers (`hugetlb` and `misc`) in a nested cgroups create scenario,
will fail because they are not, enabling the controller in the
`root_cgroup.subtree_control` file.
    
```
$ sudo cgcreate -ghugetlb:foo/bar
Error: Failed to delete foo/bar: No such file or directory
cgcreate: can't create cgroup foo/bar: No such file or directory
```
    
Fix this by enabling the controller in the `root_cgroup.subtree_control`
file unconditionally for all controllers, if not already enabled, while
calling `cgroupv2_subtree_control_recursive()` to enable the controller
in the given cgroup and its descendants. Checking and enabling every
controller unconditionally should work in the future, in case systemd
disables some other controllers too.

This patchset has two other patches, that update the `gunit/012` test
to adopt the new checks and add new `ftests` test cases that will recreate
the nested cgroups with `hugetlb` and `misc` controllers.

Fixes: #405

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -2534,6 +2534,36 @@ out:
 	return error;
 }
 
+/*
+ * Test if the controller is enabled in the root_cgrp.subtree_control file
+ * and enable the controller, if not enabled.
+ */
+static int test_and_set_ctrl_mnt_path(const char * const mount_path, const char * const ctrl_name)
+{
+	bool enabled;
+	int ret;
+
+	/* return if the controller is enabled */
+	ret = cgroupv2_get_subtree_control(mount_path, ctrl_name, &enabled);
+	if (ret == ECGOTHER)
+		return ret;
+
+	if (enabled == true)
+		return 0;
+
+	/* check if the controller is available is controllers file */
+	ret = cgroupv2_get_controllers(mount_path, ctrl_name, &enabled);
+	if (ret == ECGOTHER || ret == ECGROUPNOTMOUNTED)
+		return ret;
+
+	/* enable the controller in the subtree_control file */
+	ret = cgroupv2_subtree_control(mount_path, ctrl_name, 1);
+	if (ret)
+		return ret;
+
+	return 0;
+}
+
 /**
  * Recursively enable/disable a controller in the cgv2 subtree_control file
  *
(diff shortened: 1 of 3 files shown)


tests/ftests/090-cgcreate-non-default-controllers-v2.py
```diff
@@ -0,0 +1,85 @@
...
+from cgroup import Cgroup, Mode
+import consts
+import ftests
+import sys
+import os
+
+CONTROLLERS = ['hugetlb', 'misc']
+
+CGNAME = '090cgcreate'
+CGNAME1 = '{}/childcg'.format(CGNAME)
```
drakenclimber, line 19: nitpick - you could use `os.path.join()` here.

kamalesh-babulal: @drakenclimber Thank you so much for the review.  Sure, will use `os.path.join()` here.

tests/ftests/090-cgcreate-non-default-controllers-v2.py
```diff
@@ -0,0 +1,85 @@
...
+
+def setup(config):
+    Cgroup.create(config, CONTROLLERS, CGNAME2)
+
+
+def test(config):
+    result = consts.TEST_PASSED
+    cause = None
+
+    # checking for childcg controllers, will check in 090cgcreate.subtree_control file
```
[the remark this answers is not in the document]
kamalesh-babulal: Ah !! It sounds incorrect, how about this:
```
checking if the controller is enabled in granchildcg, in turn will check its parent cgroup childcg's subtree_control file, if it's enabled in parent cgroup, it's also enabled in the grandparent too.
```

kamalesh-babulal:
**v2 changes**:
- Addressed the review comments from @drakenclimber 
1. Used `os.path.join()` in the ftests testcase
2. Improved  the comments in the ftests testcase
- Renumbered the test case from `090` to `091`, as they are reserved for `cgset` and `cgxset` recursive test.

drakenclimber (approved):

drakenclimber:

PR #407, opened 2023-10-21
Fix gunit test build warnings

kamalesh-babulal:
This patchset fixes the build warnings across the gunit test cases,
it fixes warnings such as `unused variables`, `signed comparison`
and `string truncation` with no functional changes


kamalesh-babulal:
The `checkpatch` review fails due to c++ build warnings in the commit message, making it < 75 characters columns will make it look very hard to make any sense out of.  I don't have a strong preference, I can re-spin the patch with adhering to < 75 character columns in the commit  message.

drakenclimber (approved):

drakenclimber:

PR #420, opened 2024-03-18
feat: add cgroup_get_threads()

adriaan42:
In the same way `cgroup_get_procs()` reads the cgroup.procs file of a cgroup, the new function reads cgroup.threads, which is useful to interact with threaded cgroups.

@maintainers: Does the new function also need to be listed in `src/libcgroup.map`?

### DIFF
--- a/include/libcgroup/groups.h
+++ b/include/libcgroup/groups.h
@@ -589,6 +589,17 @@ char *cgroup_get_value_name(struct cgroup_controller *controller, int index);
 int cgroup_get_procs(char *name, char *controller, pid_t **pids, int *size);
 
 /**
+ * Get the list of threads in a cgroup. This list is guaranteed to
+ * be sorted. It is not necessary that it is unique.
+ * @param name The name of the cgroup
+ * @param controller The name of the controller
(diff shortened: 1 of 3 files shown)


kamalesh-babulal:
> Does the new function also need to be listed in src/libcgroup.map?

@adriaan42 Thanks for the patch. Yes, it needs to be added to `src/lubcgroup.map`

adriaan42:
> > Does the new function also need to be listed in src/libcgroup.map?
> 
> @adriaan42 Thanks for the patch. Yes, it needs to be added to `src/lubcgroup.map`

@kamalesh-babulal Thanks! I added a section and tried to follow the pattern already present, guessing that 3.2 might be the next release? Feel free to correct that...

kamalesh-babulal (approved):

drakenclimber (approved):

drakenclimber:
Pushed to main in commit 84ff288c1b95.  Since this is a new feature, I did not push it to release-3.1.  As discussed in the pull request, this feature will be available in libcgroup v3.2 and beyond.

Thanks, @adriaan42 and @kamalesh-babulal

PR #425, opened 2024-05-16
cgrules: Add ignore_rt option

kamalesh-babulal:
Introduce a new `ignore_rt` option for cgrules, that would ignore rules matching user:process, controller(s), destination cgroup with `SCHED_RR` and `SCHED_FF` policies.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -536,10 +536,16 @@ STATIC int cgroup_parse_rules_options(char *options, struct cgroup_rule * const 
 		cmp_len = min(strlen(stok_buff), strlen(CGRULE_OPTION_IGNORE));
 		if (strlen(stok_buff) == strlen(CGRULE_OPTION_IGNORE) &&
 		    strncmp(stok_buff, CGRULE_OPTION_IGNORE, cmp_len) == 0) {
-			rule->is_ignore = true;
+			rule->is_ignore = 1;
 			continue;
 		}
 
+		cmp_len = min(strlen(stok_buff), strlen(CGRULE_OPTION_IGNORE_RT));
+		if (strlen(stok_buff) == strlen(CGRULE_OPTION_IGNORE_RT) &&
+		    strncmp(stok_buff, CGRULE_OPTION_IGNORE_RT, cmp_len) == 0) {
+			rule->is_ignore = 2;
+			continue;
+		}
 		/*
 		 * "ignore" is the only currently supported option.
 		 * Raise an error if we get here.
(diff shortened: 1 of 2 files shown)


kamalesh-babulal:
**v2 Changes**:
- Rebased the patch to the current `HEAD`

src/api.c
```diff
@@ -536,10 +536,16 @@ STATIC int cgroup_parse_rules_options(char *options, struct cgroup_rule * const
...
 		    strncmp(stok_buff, CGRULE_OPTION_IGNORE, cmp_len) == 0) {
-			rule->is_ignore = true;
+			rule->is_ignore = 1;
 			continue;
 		}
 
+		cmp_len = min(strlen(stok_buff), strlen(CGRULE_OPTION_IGNORE_RT));
+		if (strlen(stok_buff) == strlen(CGRULE_OPTION_IGNORE_RT) &&
+		    strncmp(stok_buff, CGRULE_OPTION_IGNORE_RT, cmp_len) == 0) {
+			rule->is_ignore = 2;
```
drakenclimber, line 546: The way this is currently written, we would not be able to handle `ignore,ignore_rt` in the options field.  Is that a valid configuration?

src/api.c
```diff
@@ -536,10 +536,16 @@ STATIC int cgroup_parse_rules_options(char *options, struct cgroup_rule * const
...
+			continue;
+		}
 		/*
 		 * "ignore" is the only currently supported option.
```
drakenclimber, line 550: This comment is now out of date.  We can probably just delete it since the `cgroup_err()` below clearly states why there was a problem.

drakenclimber (approved):
I think this looks good, @kamalesh-babulal.  I have one small comment, but I don't think it's a big deal.

src/api.c
drakenclimber: If it's invalid, we should probably error out if someone has tried to set both.  If it's valid, then maybe a bit field?

kamalesh-babulal: @drakenclimber Thank you so much for the review. Agreed, I will fix the code for all three cases:
1. Only `ignore` option
2. Only `ignore_rt` option
3. Both `ignore` and `ignore_rt` options.

Sure, good idea will implement something on the same lines.

src/api.c
kamalesh-babulal: Sure, will remove the comments.

kamalesh-babulal:
**V4 Changes**:
- Rebase to current `main` branch

drakenclimber:
Pushed to main in commit b8bba7ba61fb

PR #426, opened 2024-05-17
tools/tools-common.c: handle DT_UNKNOWN from readdir()

krokodilerian:
It's possible that readdir() would return DT_UNKNOWN in d_type for some filesystems (observed on RHEL 8 with XFS and documented in readdir(3)), thus the need to do an explicit stat() for such cases.


kamalesh-babulal:
@krokodilerian Thanks for the patch. It looks good to me, with minor nitpicks of the build and checkpatch warnings:
- https://github.com/libcgroup/libcgroup/actions/runs/9127174292/job/25230628968
- https://github.com/libcgroup/libcgroup/actions/runs/9127174292/job/25230630246

can you please re-spin the patch, with the warnings fixed.

krokodilerian:
Thanks, I'll see to those.

kamalesh-babulal (approved):

kamalesh-babulal:
@krokodilerian Thanks for re-spinning the patch.

drakenclimber (approved):

drakenclimber:
`Acked-by: Tom Hromatka <tom.hromatka@oracle.com>`

PR #428, opened 2024-05-29
Introduce Kernel Version based ftests controller output map

kamalesh-babulal:
This patchset introduces Kernel version-based ftests controller
output mapping.  The first patch introduces `get_kernel_version()`
that returns the `Major.Minor.Patch Level` and the second patch
calculates the CPU controllers' expected output lines based on the
Kernel version.


drakenclimber:

PR #430, opened 2024-06-05
ftests: Refactor expected controller output

kamalesh-babulal:
This patchset refactors the expected controller(s) output across the ftests, by
moving them into `consts.py`, it also makes managing the expected output
easier.


drakenclimber (approved):

drakenclimber:

PR #434, opened 2024-06-10
ftests/cgroup: Add support for favordynmods

kamalesh-babulal:
Upstream Linux Kernel commit 81d3a5be05d3 ("cgroup: add cgroup_favordynmods=
command-line option") introduced CONFIG_CGROUP_FAVOR_DYNMODS, that would
append `favordynmods` to the mount options on cgroup v1. This patch strips that option.


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 52b196c2f4f3.  Thanks, @kamalesh-babulal

PR #436, opened 2024-06-20
ftests/071: Introduce delay before killing cgroup

kamalesh-babulal:
Fix the following issue:
`071-sudo-set_default_systemd_cgroup.py - cgroup_delete_cgroup failed: 50016`

Some older version of Kernel/Python combination, requires few jiffies to sync the
`cgroup.procs` list of process. Fix it by introducing 0.5 second delay, before removing
the cgroup.


drakenclimber (approved):

drakenclimber:
Pushed to main in commit e2da3dcc8d13.  Thanks, @kamalesh-babulal

PR #437, opened 2024-06-25
ftests: Refactor expected output iterator

kamalesh-babulal:
This patchset introduces a helper function to check if the expected output
of the controller(s) and actual output match.  Using this helper function
the test cases 009, 010, 013 are refactored to look clean and minimal.


drakenclimber (approved):

drakenclimber:

PR #439, opened 2024-07-18
Fix CodeQL warnings across code

kamalesh-babulal:
This patchset fixes two warnings reported by CodeQL about:
`Incorrect return-value check for a 'scanf'-like function`. 
The first patch fixes the warning in `tools/cgsnapshot.c` and the
second one is `src/api.c`


drakenclimber (approved):

drakenclimber:

PR #444, opened 2024-09-30
ftests/030: teach parsing of cgroup v1/v2 code isolation

kamalesh-babulal:
Starting `v6.12` cpuset and memory controllers code is split into v1 and v2 code,
where v1 code is only compiled when COFIG_CPUSET_V1 and CONFIG_MEMCG_V1
is set. Which defaults to disabled, teach the test case to ignore failures, when
the cpuset or memory controller is missing from `/proc/cgroups` on cgroup v2
booted system.


drakenclimber (approved):

drakenclimber:

Issue #441, opened 2024-09-13
Problem when moving threads in threaded cgroups


adriaan42:
I'm using cgroups v2, and have found an issue when using threaded cgroups. I'm attaching a reproducer, which does:

* Main thread creates a threaded cgroup (`/mycgroup`), and a subgroup (`mysubgroup`).
* Process is moved to `/mycgroup` (libcgroup writes the PID to `cgroup.procs`)
* `thread1` is created and moves itself to `mysubgroup` (libcgroup writes the TID to `cgroup.threads`)
* `thread1` creates `thread2`, which starts out in `mysubgroup`, but then wants to move itself back to `/mycgroup`.

This would work by writing the TID of `thread2` to `/mycgroup/cgroup.threads`, but libcgroup writes it to `/mycgroup/cgroup.procs`, resulting in ALL threads of the process being moved back to `/mycgroup`, also `thread1`.

The way I understand the problem:

* `cgroup_attach_task[_*]()` needs to determine whether to write the PID/TID to `cgroup.procs` or `cgroup.threads`.
* For `cgroup.type == domain` that's clear, because only `cgroup.procs` works
* For `cgroup.type == threaded` it's also clear, because only `cgroup.threads` works
* But for the root of the "threaded subtree" with `cgroup.type == domain threaded`, both files can be written to, and have a different meaning, either move a process with all its threads to the threaded subtree, or move a single thread of a process (that's already in the threaded subtree) to the subtree-root.

From what I can tell, even using the new `cgroup_attach_thread_tid()` (which sounds like it might do the "right" thing here), does not fix the issue.

## My program
```c
#include <errno.h>
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

#include <libcgroup.h>

#define MAINGROUP "/mycgroup"
#define SUBGROUP "/mycgroup/mysubgroup"

struct cgroup *maincg = NULL;
struct cgroup *subcg = NULL;
pthread_t thread1, thread2;

void *fn2(void *)
{
    int ret = 0;
    printf("thread2: tid=%d\n", gettid());
    printf("thread2: moving back to main group\n");
#if 0 // THIS DOES NOT WORK
    ret = cgroup_attach_task(maincg);
#endif
#if 1 // WORKAROUND
    FILE *fd = fopen("/sys/fs/cgroup" MAINGROUP "/cgroup.threads", "w");
    if (fd == NULL) {
        printf("ERROR: fopen\n");
[...]

PR #448, opened 2024-10-07
Support attaching of tid to threaded cgroup domain type

kamalesh-babulal:
This patchset supports moving the thread(s) into a threaded
cgroup domain.  Previously, given `tid` would be written into
 `cgroup.procs` invariant of attaching a task `pid` or `tid`, this
effectively moves threads sharing the same `tid` into the
thread domain.

Now with this patchset `cgroup_attach_thread_tid()` will write
into `cgroup.threads` for the threaded cgroup domain and
into `cgroup.procs` when attached from other functions.

The issue is explained in detail in #441

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -1829,7 +1829,7 @@ char *cg_build_path(const char *name, char *path, const char *type)
 }
 
 static int cgroup_get_cg_type(const char * const path, char * const type,
-			      size_t type_sz)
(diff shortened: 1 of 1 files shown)


drakenclimber:
Pushed to main.  Thanks, @kamalesh-babulal and @adriaan42

PR #454, opened 2024-12-02
github: Use default runners

drakenclimber:
Now that github actions have an image that defaults to cgroup v2, we can utilize that image rather than hosting our own github actions runners.

Use ubuntu-latest as our cgroup v2 test runner, and delete the jobs that use self-hosted runners.


kamalesh-babulal (approved):

kamalesh-babulal:
@drakenclimber Thanks for working on simplification.

kamalesh-babulal:
Pushed into main in commit https://github.com/libcgroup/libcgroup/commit/e475d7a545a2403e1ff7460bc849341838b79dd3. Thanks, @drakenclimber

PR #455, opened 2024-12-05
ftests-wrapper: remove the timeout logic

kamalesh-babulal:
With commit e475d7a545a2 ("github: Use default runners"), the dependency
on the self-hosted runners is removed. Also, remove the timeout logic to
synchronize jobs between self-hosted runners too.


kamalesh-babulal:
**V2 Changes:**
- Remove the additional stale patch.

drakenclimber (approved):

drakenclimber:
Pushed to main in commit 61f8efbdb471.  Thanks, @kamalesh-babulal

PR #456, opened 2024-12-19
cgconfig: update cgroup mount paths

kamalesh-babulal:
Update the cgroup mount paths across `cgconfig.conf*` files,
by replacing `/mnt/cgroups` with `/sys/fs/cgroup/`


drakenclimber:
Pushed to main in commit 2adf17c8ff79.  Thanks, @kamalesh-babulal

PR #457, opened 2024-12-20
ftests/consts.py: Add more cpu controller output (v2)

kamalesh-babulal:
Upstream v6.13 kernel commit aefa398d93d5 ("cgroup/rstat: Tracking
cgroup-level niced CPU time") added "nice_usec" to the cgroup base
stats cputime, add it to the list of cpu controller (v2) expected output.


drakenclimber:

PR #460, opened 2024-12-27
api: Show all members of a given group

aarontomlin:
In the context of a group rule (i.e. indicated by '@' used to prefix the actual group name), getgrnam(3) is used to provide a pointer to a group file entry that may contain a NULL-terminated array of pointers to group members. A user can belong to multiple groups. With this information, we then check the username that corresponds to the specified UID against each group member for a match. This patch makes it possible to see this information if debug level logging is enabled.

For example, this can be useful when troubleshooting the Cgroup rules engine daemon.

Signed-off-by: Aaron Tomlin <atomlin@atomlin.com>

PR #463, opened 2025-01-15
api: Show all members of a given group

drakenclimber:
I've made some small changes to pull request #460 in an attempt to make it better match libcgroup's coding style.  It should behave the same as the original pull request.


kamalesh-babulal (approved):

drakenclimber:
Pushed to main.  Thanks, @aarontomlin for the patch, and thanks @kamalesh-babulal for the review.

FYI - @aarontomlin, I made sure that you were the author for 9155a813ff18 since you wrote the vast majority of that patch.  Thanks so much for the submission

aarontomlin:
My pleasure.

PR #465, opened 2025-01-24
src/config: Fix data race reported by Coverity

kamalesh-babulal:
Fix the following data race issue reported by Coverity:

CID 465888:Check of thread-shared field evades lock acquisition
(LOCK_EVASION):

```
The data guarded by this critical section may be read while in an
inconsistent state or modified by multiple racing threads.

In cgroup_config_insert_into_namespace_table: Checking the value
of a thread-shared field outside of a locked region to determine if a
locked operation involving that thread shared field has completed.
```

Fix it by moving the `namespace_table_index` value check too under the 
`namespace_table_lock`

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -599,10 +599,12 @@ int cgroup_config_insert_into_namespace_table(char *name, char *nspath)
 {
 	char *ns_tbl_name, *ns_tbl_path;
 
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 322e81c20f44.  Thanks, @kamalesh-babulal

PR #469, opened 2025-01-31
 Update libcgroup.h to make systemd optional.

nater1983:
Update libcgroup.h to make systemd optional.

### DIFF
--- a/include/libcgroup.h
+++ b/include/libcgroup.h
@@ -18,7 +18,10 @@
 #include <libcgroup/config.h>
 #include <libcgroup/log.h>
 #include <libcgroup/tools.h>
+
+#ifdef WITH_SYSTEMD
 #include <libcgroup/systemd.h>
+#endif
 
 #undef _LIBCGROUP_H_INSIDE
 


kamalesh-babulal:
Thanks for the patch @nater1983 can you please why you want to `ifdef` `systemd.h` file.

nater1983:
Because if you do —enable-systemd=no then systemd.h shouldn’t be installed
by default.

Nathaniel & LeAnna Russell

drakenclimber:
There are two common ways to completely enable/disable a feature in a C program:
1. Entirely surround it with `#ifdef`
2. Create stub functions that provide the same API as if the feature were enabled, but the stub functions obviously do nothing

In libcgroup, we've opted for the second solution for a variety of reasons:
* The Linux kernel does it this way for most features, and we want to emulate that model in libcgroup
* Automated testing
  * all of the tests will run with or without systemd.  We carefully wrote the systemd code and tests so that they'll behave properly in both cases
  * If we switch to `#ifdef`, we would have to skip tests in non-systemd builds.  Our testing framework supports this, but it would take effort
* The byacc logic in libcgroup is systemd-aware.  Putting `#ifdef` in `parse.y` would work, but is non-standard and unexpected

If you look at our [continuous integration](https://github.com/libcgroup/libcgroup/actions), you'll see that we build libcgroup without systemd defined.  The build and tests all pass successfully.

@nater1983 - What exactly are you trying to do?  And why does the current libcgroup solution not meet your needs?  I'm not necessarily against a change like this, but as you can see from our automated testing, it breaks a lot of things.  We would need a _really_ good reason to make such a switch.

nater1983:
I understand that but for Slackware we had to put a blank stub file in its
place so that other programs can still compile.

Nathaniel & LeAnna Russell

nater1983:
http://ftp.slackware.com/pub/slackware/slackware64-current/source/a/libcgroup/libcgroup.SlackBuild


Nathaniel & LeAnna Russell

drakenclimber:
Ahhh... I see.  It looks like your issue has been fixed by this commit https://github.com/drakenclimber/libcgroup/commit/592dcdcf243576bd2517d3da9bc18990de08e37e.

FYI - We're planning on doing a new release in the next month or so.  It will obviously include this change.

nater1983:
I gotcha i didnt realize there hasnt been a release in 2 yrs. That makes sense i will forqard the patch off to Pat thank you.

PR #471, opened 2025-02-18
all: Add THIRD_PARTY_LICENSES.txt

drakenclimber:


kamalesh-babulal:
@drakenclimber The patch looks good to me, thanks for including it.  There are few `checkpatch` warnings those can be classified as  `inclusion of address` which can be ignored and other is trivial `trailing whitespace`, warnings. I was able to find the license in html format only, and I am guessing that copy, paste from the html page would have resulted in whitespaces at beginning of few blank lines. Can we please fix it, to stay consistent across the License text.

kamalesh-babulal (approved):

drakenclimber:
Pushed to main in commit 56a6ef4dff01.  Thanks, @kamalesh-babulal

PR #473, opened 2025-03-27
doc: Add documentation for cgrules' options

drakenclimber:
Add man page documentation for the options field in cgrules configuration files.  Currently only "ignore" and "ignore_rt" are supported.


kamalesh-babulal (approved):

drakenclimber:
Pushed to main in commit 4ce8819b606f.  Thanks for the review, @kamalesh-babulal

PR #476, opened 2025-04-01
Pass -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to CMake

kamalesh-babulal:
Googletest is broken across the Ubuntu images provided by GitHub,
and it complains about incompatibility with CMake versions < 3.5.
Since we use googletest as a third-party submodule, which we do
not have control over, lets pass `-DCMAKE_POLICY_VERSION_MINIMUM=3.5`
to keep CMake compatible as a workaround.


kamalesh-babulal:
#477 bumps the googletest version to 1.2.0 and gtest are successful.

PR #477, opened 2025-04-01
Various continuous integration fixes

drakenclimber:
* update to a newer version of googletest
* fix erroneous string usage in two unit tests
* test 089 fails on some containers - skip it
* retry getting the sleep pid in ftests.  it sometimes fails


kamalesh-babulal (approved):

drakenclimber:
Pushed to main.  Thanks for the review, @kamalesh-babulal

PR #479, opened 2025-04-08
src/daemon/cgrulesengd: Correctly check for ENOBUFS when doing recvfr…

freemangordon:
…om()

Currently, code checks if recvfrom() returns ENOBUFS, however, it can never return errno type error code, but -1 on error

Fix that by checking errno in case of an error instead


drakenclimber:
Pushed to main in commit aaf329d63056.  Thanks, @freemangordon

drakenclimber:
Fixes commit 2d10aa26dc974 ("Cleanup: Separate the netlink method to cgre_receive_netlink_msg().")

PR #480, opened 2025-04-09
Rework testing and how they're run during distro builds

drakenclimber:
Now that Github Actions don't support cgroup v1, we can make the github
runners run the tests in a more granular fashion.  Add a runner to
individually run the containerized tests, the non-containerized tests,
and the sudo tests.  This simplifies the testing logic, and should help
us better identify failures.

As part of this, no longer run ftests-wrapper.sh when `make check` is
invoked on the ftests directory.  This is an undue burden for distros
as it requires uncommon dependencies (lxc, etc.) and can adversely
affect the cgroup sysfs filesystem on the build system.  The onus is
on upstream libcgroup to provide a well-tested and vetted *.tar.gz for
the distros that they can then just package up and disseminate.  They
shouldn't have to deal with challenging tests and test configurations.

Finally, add a configure option to enable/disable the unit tests.
Because of their dependency upon googletest, they're very difficult to
run for distros.  This allows them to packaged with libcgroup, but they
are disabled by default.  They can be enabled via the config option
`--enable-unittests`.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -128,27 +128,6 @@ jobs:
         name: Unit Test Code Coverage
         path: lcov.*
 
-  distcheckv1:
-    name: Cgroup v1 Dist Check
-    runs-on: ubuntu-20.04
-
-    steps:
-    - name: Install container dependencies
-      run: |
-        sudo apt-get update
-        sudo apt-get install lxc lxd-installer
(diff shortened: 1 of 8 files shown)


kamalesh-babulal (approved):

kamalesh-babulal:
@drakenclimber Thank you so much for working on it. Pushed to main

PR #481, opened 2025-04-15
Memory abstraction layer

drakenclimber:


kamalesh-babulal:
@drakenclimber Thanks for working on adding the memory abstraction. It looks great overall. I have a few minor comments on the patch.

src/abstraction-memory.c
```diff
@@ -0,0 +1,104 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+#include <stdio.h>
+
+#define V1_MAX_STR "9223372036854771712"
+#define V2_MAX_STR "max"
+
+#define LL_MAX 8192
```
kamalesh-babulal, line 24: The maximum allowed value range is of 20 characters and you may want to reduce the array size here to reduce the unused bytes.

src/abstraction-memory.c
```diff
@@ -0,0 +1,104 @@
...
+		const char * const in_setting, const char * const in_value,
+		const char * const out_setting)
+{
+	char max_line[LL_MAX] = {0};
+	int ret;
+
+	if (strlen(in_value) == 0) {
+		ret = cgroup_add_value_string(dst_cgc, out_setting, NULL);
+		if (ret)
+			goto out;
```
kamalesh-babulal, line 36: Its coding style preference comment, you may want to remove the check for `ret` and goto branch. They are nop, as the next statement in both if `ret` checks will be `return ret` but without the extra branch and check.

drakenclimber: I think this is a good idea here.  It cleans it up a lot and the `out` tag can be entirely removed.

src/abstraction-memory.c
```diff
@@ -0,0 +1,104 @@
...
+static int convert_max_to_limit(struct cgroup_controller * const dst_cgc,
+		const char * const in_setting,  const char * const in_value,
+		const char * const out_setting)
+{
+	char limit_line[LL_MAX] = {0};
+	int ret;
+
+	if (strlen(in_value) == 0) {
+		ret = cgroup_add_value_string(dst_cgc, out_setting, NULL);
+		if (ret)
```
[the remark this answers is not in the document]
drakenclimber: Done :+1:

src/abstraction-memory.c
```diff
@@ -0,0 +1,104 @@
...
+	char max_line[LL_MAX] = {0};
+	int ret;
+
+	if (strlen(in_value) == 0) {
+		ret = cgroup_add_value_string(dst_cgc, out_setting, NULL);
+		if (ret)
+			goto out;
+	} else {
+		if (strcmp(in_value, V1_MAX_STR) == 0)
+			snprintf(max_line, LL_MAX, "%s", V2_MAX_STR);
```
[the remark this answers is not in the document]
drakenclimber: Good find.  Totally forgot about `-1`

src/abstraction-memory.c
```diff
@@ -0,0 +1,97 @@
...
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+#include <stdio.h>
+
+#define V1_NEG1_STR "-1"
+#define V1_MAX_STR "9223372036854771712"
+#define V2_MAX_STR "max"
+
+#define LL_MAX 24
```
kamalesh-babulal, line 25: @drakenclimber Should `LL_MAX` match upper limit allowed by Kernel of 21 characters (including the '\0')? The Kernel would silently trim the excess trailing digits and update setting with `V1_MAX_STR` value but this will leave the user wondering what went wrong. Either we can leave there, mimicing the Kernel behavior or we would check the strlen and return error?

drakenclimber: Sounds good to em.  I'll make that change

tests/ftests/093-cgxget-memory_settings.py
```diff
@@ -0,0 +1,169 @@
...
+            cause = (
+                        'After setting {}={}, expected {}={}, but received '
+                        '{}={}'
+                        ''.format(entry[0], entry[1], entry[3], entry[4],
+                                  entry[3], out)
+                    )
+            return result, cause
+
+    return result, cause
+
```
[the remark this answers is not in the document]
drakenclimber: Good idea.  I'll add it.

kamalesh-babulal (approved):

kamalesh-babulal:
@drakenclimber Thanks for working on the mm-abstraction. Pushed into main.

PR #485, opened 2025-05-22
tests: fix checking wrong result in 017-API_fuzz_test

thesamesam:
There seem to be various calls to cgroup_add_controller where the result isn't validated, instead checking the cgrp pointer.

We could add two asserts in each place being changed here if the concern was testing for if cgroup_add_controller mangled cgrp, but it looks more likely that it's a copy/paste error instead and we're just testing the wrong thing.


kamalesh-babulal (approved):

drakenclimber (approved):

drakenclimber:
Thanks for the patch, @thesamesam

Pushed to main in commit 86a371889ca7.  Pushed to release-3.2 in commit 32eff99aeefc.

PR #499, opened 2025-07-08
ftests: Fix false positive in 010 testcase

kamalesh-babulal:
Fix the false positive in the testcase number 010, the return value should
be by default `TEST_FAILED`, instead of `TEST_PASSED`. The flow of the
testcase is designed to set the return value to `TEST_PASSED` only if the
current kernel's cpu controller settings matches any of the expected
cpu controller output list.


kamalesh-babulal:
This patch is dependent on #501

drakenclimber:
Pushed to the main and the release-3.2 branches since this will allow the tests to run on newer kernels.

PR #502, opened 2025-07-10
ftests/cgroup: Add timeout to cgrulesengd

kamalesh-babulal:
`cgrulesengd -n` (nodaemon mode) runs indefinitely until it is terminated
with `SIGKILL`. To handle this, introduce a 20-second timeout that is passed
to `__run()`, which in turn calls `subproc.kill()` to terminate `cgrulesengd`
once the timer expires.

### DIFF
--- a/tests/ftests/cgroup.py
+++ b/tests/ftests/cgroup.py
@@ -705,7 +705,7 @@ class Cgroup(object):
                                 'supported'
                             )
         else:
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 5455ce91abc0 and release-3.2 in commit ec082fe9ad3e

Issue #505, opened 2025-07-28
pointer-overflow yylex() src/lex.l:47:69


WhiteWLf-dev:
If I give the function my data:
```
#include "config.h"
#include "libcgroup.h"
#include "libcgroup-internal.h"
#include "parse.h"

extern int cgroup_init_templates_cache(char *pathname);

int main(int argc, char **argv)
{
  cgroup_init_templates_cache(argv[1]);
  return (0);
}
```
I got the UBSAN report:
```
    lex.l:47:34: runtime error: addition of unsigned offset to 0x555556296510 overflowed to 0x55555629650f
        #0 0x5555555b0d43 in yylex /libcgroup-UBSAN/src/lex.l:47:69
        #1 0x5555555a9073 in yyparse /libcgroup-UBSAN/src/parse.c:1199:16
        #2 0x5555555a8a66 in cgroup_parse_config /libcgroup-UBSAN/src/config.c:1084:9
        #3 0x5555555a4ee3 in cgroup_init_templates_cache /libcgroup-UBSAN/src/config.c:1578:8
        #4 0x5555555a4ee3 in main /libcgroup-UBSAN/templates-cache.c:12:3
        #5 0x7ffff7cf2249 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
        #6 0x7ffff7cf2304 in __libc_start_main csu/../csu/libc-start.c:360:3
        #7 0x555555575b80 in _start (/out/UBSAN/templates-cache+0x21b80) (BuildId: 853a0c47d081e41b3784dda0f7f38d4aec1f5dce)
    
    SUMMARY: UndefinedBehaviorSanitizer: pointer-overflow lex.l:47:34
```
The error is related to overflow when adding an unsigned offset when processing quoted strings.

        43     \"namespace\"\t{return NAMESPACE;}
        44     \"template\"\t{return TEMPLATE;}
        45     \"default\"\t{yylval.name = strdup(yytext); return DEFAULT;}
        46     [a-zA-Z0-9_\\-\\/\\.\\,\\%\\@\\\\]+ {yylval.name = strdup(yytext); return ID;}
    --->47     \\\"[^\"]*\\\" {yylval.name = strdup(yytext+1); yylval.name[strlen(yylval.name)-1] = '\\0'; return ID; }
        48     .\t{return yytext[0];}
        49     %%
        50     
[...]

PR #506, opened 2025-07-28
Fix Issue #505

WhiteWLf-dev:
The error is related to overflow when adding an unsigned offset when processing quoted strings.

    43     \"namespace\"\t{return NAMESPACE;}
    44     \"template\"\t{return TEMPLATE;}
    45     \"default\"\t{yylval.name = strdup(yytext); return DEFAULT;}
    46     [a-zA-Z0-9_\\-\\/\\.\\,\\%\\@\\\\]+ {yylval.name = strdup(yytext); return ID;}
--->47     \\\"[^\"]*\\\" {yylval.name = strdup(yytext+1); yylval.name[strlen(yylval.name)-1] = '\\0'; return ID; }
    48     .\t{return yytext[0];}
    49     %%
    50     
If the string is empty (""), then strlen(yylval.name -1 can give a negative value converted to a very large unsigned number.

### DIFF
--- a/src/lex.l
+++ b/src/lex.l
@@ -39,7 +39,15 @@ jmp_buf parser_error_env;
 "systemd"     	{return SYSTEMD;}
 "default"	{yylval.name = strdup(yytext); return DEFAULT;}
 [a-zA-Z0-9_\-\/\.\,\%\@\\]+ {yylval.name = strdup(yytext); return ID;}
-\"[^"]*\" {yylval.name = strdup(yytext+1); yylval.name[strlen(yylval.name)-1] = '\0'; return ID; }
+\"[^"]*\"	{
+	if (yytext[0] != '\0' && yytext[1] != '\0') {
+		yylval.name = strdup(yytext+1); 
+		yylval.name[strlen(yylval.name)-1] = '\0';
+	} else {
+		yylval.name = strdup("");
+	} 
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
@WhiteWLf-dev Thank you for trying it on clang and fixing the issue. From the #505 the template example file https://github.com/user-attachments/files/21466644/myfilelibcgroup.txt seems to be corrupted?

A few nit picks on the patch:
- We follow the Linux Kernel contribution guidelines and the patch would require your `signed-off`. 
- There are few checkpatch issue, can you please fix them

WhiteWLf-dev:
@kamalesh-babulal The file was generated by fuzzer (and the code from example is harness for the fuzzer). And the file wasn't corrupted.

kamalesh-babulal:
@WhiteWLf-dev I'm curious about the fuzzer you have been using. Can you share more details about it?

kamalesh-babulal:
@WhiteWLf-dev Thanks for the patch, push in to main https://github.com/libcgroup/libcgroup/commit/05ce62bca993c260af6478a1f2035bb0c73050a9 and release-3.2 commit https://github.com/libcgroup/libcgroup/commit/fe34851e64ecce88d0e936394e79dc06f66d3b9b

Issue #497, opened 2025-07-03
potential issue with CGRP_RULE_MAXLINE


grey3228:
Hello!

I'd like to clarify my concern about `CGRP_RULE_MAXLINE` constant. It is defined like below:

```
#define CGRP_RULE_MAXLINE	(FILENAME_MAX + CGRP_RULE_MAXKEY + CG_CONTROLLER_MAX + 3)
```

As I understand, there was added a new "options" field for cgrule in [fbb0489a4780258aa577b9d08274eb0772292d70](https://github.com/libcgroup/libcgroup/commit/fbb0489a4780258aa577b9d08274eb0772292d70). But for some reason value of CGRP_RULE_MAXLINE wasn't changed. 

If I understand correctly, it may cause problems in function src/api.c/cgroup_parse_rules_file, where size of `buff` for reading lines of cgrules conf file is exactly `CGRP_RULE_MAXLINE` and there is corresponding `options` buffer, which obviously must hold value of cgrule options, and it is actually used for reading options from `buff` (`itr`). 

Do you think `CGRP_RULE_MAXLINE` should be modified to account options (which size is `CG_OPTIONS_MAX`)? For example, like that:

```
#define CGRP_RULE_MAXLINE	(FILENAME_MAX + CGRP_RULE_MAXKEY + CG_CONTROLLER_MAX + CG_OPTIONS_MAX + 4)
```

Thank you for your time and expertise!

kamalesh-babulal:
@grey3228 Agreed ! you're technically right. Could you please open a PR with your suggestion as a patch?

kamalesh-babulal:
Merged the patch from #508 that fixes this issue. Closing the issue.

PR #508, opened 2025-07-31
fix: make CGRP_RULE_MAXLINE include CG_OPTIONS_MAX

grey3228:
CGRP_RULE_MAXLINE const, which is used as a length for buffer for parsed cgroup rules strings, doesn't include length of options field (CG_OPTIONS_MAX).

Include CG_OPTIONS_MAX in CGRP_RULE_MAXLINE to avoid potential buffer overflow in a process of parsing rules.

Fixes: #497


kamalesh-babulal:
@grey3228 Thanks for the patch. Pushed into main commit https://github.com/libcgroup/libcgroup/commit/88fc455b8d0df20910ee2f8e90eb90a4752bd986 and release-3.2 in commit https://github.com/libcgroup/libcgroup/commit/041a129199de28731c439ad425e63e084fdc3713

PR #513, opened 2025-11-19
src/api.c: prevent potential oom null derefs

grey3228:
Hello!

I used SAST tool Svace to analyze source code and encountered 2 potential null ptr dereference cases:

1) In function `cgroup_fill_cgc` there is an allocation of memory by calling `strdup(ctrl_dir->d_name)`. Result of this allocation is not checked and used in `strcmp` then;
2) In function `cg_get_cgroups_from_proc_cgroups` there is an allocation of memory by calling `malloc(buff_len)`. Result of this allocation is not checked and then used in `snprintf`.

This PR adds handling of such possible failed memory allocations.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -5930,6 +5930,11 @@ STATIC int cg_get_cgroups_from_proc_cgroups(pid_t pid, char *cgrp_list[],
 		if (buff_len > 1) {
 			/* Strip off the leading '/' for every cgroup but the root cgroup */
 			cgrp_list[idx] = malloc(buff_len);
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commits 4938e7e5d0a6 and 8a40bf6f6f99.  Pushed to the release-3.2 branch in commits 29f7e2c256a5 and 14a1e734c6d1.

Thanks for the fix and thanks for your patience, @grey3228

PR #515, opened 2026-01-14
systemd: expand hyphenated slice names

kamalesh-babulal:
When systemd receives a slice name containing one or more '-' characters,
it treats it specially and creates a nested cgroup hierarchy: it takes
the prefix up to the first '-', appends ".slice", then repeats for each
subsequent segment of the slice name, until the leaf slice name matches
the original input.

We create slices/scopes via sd-bus APIs, where systemd performs this
expansion. However, the cgroup_create_scope2() later passes the
unexpanded slice/scope name to cgroup_create_cgroup() for ownership
changes, which can result in creating a slice/scope under an
unintended, unexpanded path.

Fix by introducing cgroup_exp_slice_hyphen_name(), which expands the
slice name to the expected systemd hierarchy before passing it to
cgroup_create_cgroup() and applying ownership changes.

This patchset also adds a test case to exercise the slice name expansion
introduced by cgroup_exp_slice_hyphen_name().

### DIFF
--- a/tests/ftests/095-sudo-cgcreate_default_system_scope_hyphen.py
+++ b/tests/ftests/095-sudo-cgcreate_default_system_scope_hyphen.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: LGPL-2.1-only
+#
+# Test to create a default systemd scope with hyphen using cgcreate
+#
+# Heavily adapted from 079-sudo-cgcreate_default_systemd_scope.py
+#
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
**V3 Changes**
- Add checks to prevent potential buffer overruns in the scope name expansion function.

**V2 Changes**
- Fix the missing systemd default cgroup path when the slice/scope path is expanded.
- Simplify the expansion logic.
- Add a test case to validate the systemd default cgroup path for an expanded slice/scope path.

PR #536, opened 2026-05-05
fix potential template_table realloc mem leak

grey3228:
Hello!

In `cgroup_expand_template_table` `realloc()` leaves the original allocation untouched on failure, but the current code overwrites the only pointer to the existing `template_table` with `NULL`. This leaks the old cache and leaves global template state in an inconsistent state.

Store the result in a temporary pointer and only update `template_table` after a successful reallocation.

Found by Linux Verification Center (linuxtesting.org) with SVACE.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1809,13 +1809,16 @@ int cgroup_add_cgroup_templates(int offset)
  */
 int cgroup_expand_template_table(void)
 {
+	struct cgroup *new_template_table;
 	int i;
 
-	template_table = realloc(template_table,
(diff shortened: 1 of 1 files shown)


grey3228:
Hi @drakenclimber and @kamalesh-babulal, gentle ping on this fix.

The PR has been open since May, is conflict-free, and all CI checks pass. The realloc failure is not limited to process shutdown: it can happen during template-cache reload in a long-running cgrulesengd process, leaking the old cache and leaving template_table inconsistent with template_table_index.

Could you please take a look when you have a chance?

PR #538, opened 2026-05-13
src/libcgroup-internal: drop dirent/mntent includes

kamalesh-babulal:
src/libcgroup-internal.h includes <dirent.h> and <mntent.h>, but only uses pointer
types in function declarations.

Replace the heavy includes with forward declarations:
- add struct dirent;
- add struct mntent;
- remove <dirent.h> and <mntent.h> from the internal header

This trims transitive header dependencies and reduces compile noise, with no
functional change.

Build tested with:
  make -j4

### DIFF
--- a/src/libcgroup-internal.h
+++ b/src/libcgroup-internal.h
@@ -18,14 +18,15 @@ extern "C" {
 #include <libcgroup.h>
 
 #include <pthread.h>
-#include <dirent.h>
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 5629eb4db8cc930c9f38ee3028522fe026969ccf.  Thanks, @kamalesh-babulal

PR #539, opened 2026-05-23
abstraction/tests: preserve cpu burst mappings across v1<->v2 conversion

kamalesh-babulal:
This series preserves CPU burst settings across cgroup v1<->v2 conversion.

Patch 1 adds symmetric abstraction mappings:
```
cpu.cfs_burst_us -> cpu.max.burst
cpu.max.burst -> cpu.cfs_burst_us
```
Patch 2 extends ftests `069/070` to validate both directions. The mapping
is name-only (no scaling), since both settings represent CFS burst runtime
in microseconds. Tests mark burst rows optional so older kernels without
burst files are skipped cleanly.

### DIFF
--- a/src/abstraction-map.c
+++ b/src/abstraction-map.c
@@ -25,6 +25,8 @@ const struct cgroup_abstraction_map cgroup_v1_to_v2_map[] = {
 	{cgroup_convert_cpu_quota_to_max, "cpu.cfs_quota_us", NULL, "cpu.max", NULL},
 	{cgroup_convert_cpu_period_to_max, "cpu.cfs_period_us", NULL, "cpu.max", NULL},
 	{cgroup_convert_passthrough, "cpu.idle", NULL, "cpu.idle", NULL},
+	{cgroup_convert_passthrough, "cpu.uclamp.min", NULL, "cpu.uclamp.min", NULL},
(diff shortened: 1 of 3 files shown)


kamalesh-babulal:
**V2 Changes**
Appended 4 patches to extend CPU abstraction coverage and test validation:

- 1/4 abstraction: preserve cpu.idle across v1/v2 conversion
   Add symmetric passthrough mapping for cpu.idle in both abstraction tables
   so cgroup v1/v2 conversion retains explicit idle policy.
- 2/4 tests/ftests: cover cpu.idle passthrough abstraction
  - Extend 069/070 sudo cgxset/cgxget conversion matrices with cpu.idle
    read/write cases in both v1->v2 and v2->v1 paths.
   - Mark rows optional, skip only known unsupported-setting failures.
- 3/4 abstraction: add cpu.uclamp.min/max mappings
  - Add symmetric passthrough mappings for cpu.uclamp.min and
    cpu.uclamp.max in both abstraction tables.
  - Preserve utilization-clamp performance policy across conversions.
- 4/4 tests/ftests: cover cpu.uclamp.min/max passthrough abstractions
  - Extend 069/070 with uclamp min/max conversion checks for both
    cgroup v1/v2.
  - Keep rows optional for kernels that do not expose uclamp files.

drakenclimber (approved):

drakenclimber:

